PR #3, opened 2019-10-03
Added jekyll configs and fixed some formatting.

antaresc:
- Added Gemfiles to enable local builds
- Added README
- Added gitignore
- Fixed formatting issues with Python args


_config.yml
```diff
@@ -1,6 +1,7 @@
-remote_theme: pmarsceill/just-the-docs
+theme: "just-the-docs"
+# remote_theme: pmarsceill/just-the-docs
```
shifucun: remove this?

_config.yml
```diff
@@ -1,6 +1,7 @@
...
 
 # logo: "/assets/dc.png"
```
shifucun: can remove this as well

api/python/api_key.md
```diff
@@ -22,4 +22,4 @@ for more details.
 
 **Arguments**
 
-*   `api_key (str)` - The API key.
+*   `api_key` (`str`): The API key.
```
beets: For these -- I think it's cleaner not to escape the 'code' style multiple times (see tjann's comment on the group thread).

api/python/query.md
```diff
@@ -50,8 +50,6 @@ Be sure to initialize the library, and specify the API key. Check the [Python li
 
 ## Examples
 
-**Examples**
-
```
beets: thanks for fixing this

PR #19, opened 2020-05-19
Upper case Sheets in title and add "Google"

tjann:
to disambiguate sidebar.

### DIFF
--- a/api/sheets/get_dcid.md
+++ b/api/sheets/get_dcid.md
@@ -1,8 +1,8 @@
 ---
 layout: default
-title: Get place DCID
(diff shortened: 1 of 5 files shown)


mvashishtha:
Thank you. I think you also need to change the "parent" in all of the files in https://raw.githubusercontent.com/datacommonsorg/docsite/master/api/sheets/

tjann:
> Thank you. I think you also need to change the "parent" in all of the files in https://raw.githubusercontent.com/datacommonsorg/docsite/master/api/sheets/

Done

PR #36, opened 2020-07-10
Check in autogenerated human readable StatVar ID list.

tjann:


tjann:
This file was generated via a python script that parsed the latest version of the autogenerated StatVar MCF. The code currently lives inside google's internal codebase. The plan is for future iterations to generate the StatVar MCF and this readme file at the same time.

PR #43, opened 2020-07-19
Built jekyll theme for docs

beets:
Update styles to match datacommons.org 
* Built new jekyll theme based on https://github.com/pmarsceill/just-the-docs
* Add code syntax highlighting theme from https://raw.githubusercontent.com/jwarby/pygments-css/master/friendly.css
* Fix sidebar ordering
* Added bootstrap 4.0.0 (has to be included in the _sass folder)


beets:
Thanks! We can enforce https after this is in, since we don't load external sources anymore.

shifucun:
this is very cool! Is there a way to reference the bootstrap css from somewhere? In current way, might need some license headers in those files? There might be some requirement for included 3rd party code: go/patching#license-headers-and-copyright-notices

beets:
Yea, good point. Will add that in a follow up. There doesn't seem to be a good way of including it another way while still letting github pages to compile and deploy the pages (other methods require local builds and deploying that instead)

_config.yml
```diff
@@ -1,6 +1,39 @@
...
-# logo: "/assets/dc.png"
-title: "datacommons.org"
+# Site settings
+# These are used to personalize your new site. If you look in the HTML files,
+# you will see them accessed via {{ site.title }}, {{ site.email }}, and so on.
+# You can create any custom variable you would like, and they will be accessible
+# in the templates via {{ site.myvariable }}.
+title: "Data Commons Docs"
+description: Data Commons Documentation
+url: "https://docs.dataacommons.org"
```
mvashishtha, line 18: @beets I think this typo breaks some links, e.g. the one to "Statistical Variables" on [this page](https://docs.datacommons.org/api/).

beets: Thanks for pointing it out. Fixed in #46

PR #45, opened 2020-07-20
Misc updates

beets:
- Add licensing info for root and bootstrap
- Move bootstrap to third_party (still can't get it working from a top-level third_party directory though)
- Move sheets related images to assets/images/sheets
- Update gem packages to fix security alert on activesupport
- Fix base URL typo and use relative URLs in the sidebar


_config.yml
```diff
@@ -15,7 +15,7 @@
 # in the templates via {{ site.myvariable }}.
 title: "Data Commons Docs"
 description: Data Commons Documentation
-url: "https://docs.dataacommons.org"
+url: "https://docs.datacommons.org"
```
tjann, line 18: Nice catch

PR #53, opened 2020-08-24
Python documentation for get_stat_value()

shifucun:


api/python/stat_value.md
```diff
@@ -0,0 +1,58 @@
...
+
+* `observation_period (str)`: (Optional) The preferred `observationPeriod` for the `stat_var`.
+
+* `unit (str)`: (Optional) The `dcid` of the preferred `unit` for the `stat_var`.
+
+* `scaling_factor (str)`: (Optional) The preferred `scalingFactor` for the `stat_var`.
+
+**Returns**
+
+ A `float` value of the `stat_var` for `place`, filtered by the optional parameters. If the optional parameters are specified, but are unavailable for the `stat_var`, then ... is returned.
```
beets, line 39: please fill the ___ in, and an example of when it's overspecified (with and without possible results).

shifucun: confirmed with tjann,  if there is no value found, an ValueError will be thrown. Updated this part.

shifucun:
> Thanks! Can we please add an example for an over-constrained call?

Added example

PR #54, opened 2020-08-25
Add python api documentation for get_stat_series()

shifucun:
Also move the api_key section to as the first child under "Python" so adding new api does not require chaning the order.

### DIFF
--- a/api/python/stat_value.md
+++ b/api/python/stat_value.md
@@ -8,7 +8,7 @@ grand_parent: API
 
 # Get Statistics Value for a Place.
 
-## `get_stat_value(place, stat_var, date=None, measurement_method=None,observation_period=None, unit=None, scaling_factor=None)`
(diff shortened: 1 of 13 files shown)


api/python/stat_value.md
```diff
@@ -60,7 +60,7 @@ In next example, the parameter "observation=P3Y" overly constrains the request s
 throws ValueError:
 
 ```python
->>> get_stat_series('geoId/06085', 'Count_Person', observation_period='P3Y')
+>>> get_stat_value('geoId/06085', 'Count_Person', observation_period='P3Y')
```
beets, line 63: should we also use datacommons.get_stat_value here?

shifucun: right, needs the import statement in these examples. I will correct them in next PR

shifucun:
thanks for review!

PR #59, opened 2020-08-26
Unlist old get_stats API, and add vscode to gitignore.

tjann:


tjann:
Visual proof of hiding:

Gone from REST and Python

image

Not moved to bottom (truly hidden):
image

PR #83, opened 2020-10-13
Add REST and Python examples of API calls with usage

beets:
For REST calls, embed JSFiddle's with example usage (account details on valentine)
Update Python example result for get_stat_all


shifucun (approved):
very cool!

beets:
thanks!

PR #88, opened 2020-10-26
Add documentation for sheets sorting with video illustration

shifucun:


api/sheets/sort.html
```diff
@@ -0,0 +1,25 @@
...
+
+<h1>How To Sort</h1>
+
+<p>To sort columns of data from the API, we suggest to use the "filtered view" in
+Google Sheet. The steps are:</p>
+
+<ul>
+<li>Populate an array formula using  the {"HEADER"; formula} syntax.</li>
+<li>Turn on filter views (Data > Filter views).</li>
+<li>Click the reverse triangle on each column to sort by that column (and exclude row 1 header).</li>
```
tjann, line 17: There's a sneaky extra space after "using"?

Maybe I am just not familiar with Sheets API, but from reading this, I am not sure what the first bullet point means, or what exclude row I header means. If normal user also has no sense, I wonder if it's better to move this under the video as a "summary" (can then assume user has seen the video)

tjann: Just watched the video. I think it's worth hyper linking to the {"HEADER"; formula}.

It looks like the exclusion of row 1 header was automatic. Is that correct? Maybe we can omit that in this doc then? Otherwise It seemed like I needed to do something to get the header excluded.

tjann (approved):

api/sheets/sort.html
shifucun: Good suggestions, explicitly calling this out does create more questions. I moved this down the video and omit the first step.

api/sheets/sort.html
```diff
@@ -0,0 +1,25 @@
...
+layout: default
+title: How To Sort
+nav_order: 8
+parent: Google Sheets
+grand_parent: API
+---
+
+<h1>How To Sort</h1>
+
+<p>To sort columns of data from the API, we suggest to use the "filtered view" in
```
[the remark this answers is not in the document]
shifucun: Done

api/sheets/sort.html
```diff
@@ -0,0 +1,25 @@
+---
+layout: default
+title: How To Sort
+nav_order: 8
+parent: Google Sheets
+grand_parent: API
+---
+
+<h1>How To Sort</h1>
```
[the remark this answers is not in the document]
shifucun: Done

PR #89, opened 2020-11-01
Rest docs overhaul

KilimAnnejaro:
- Removed API key documentation from general REST page
- Used Redoc for documentation of triples and SPARQL endpoints (moved away from Swagger)
- Unpublished current markdown doc for triples endpoint (wanted to retain the content as a backup)

Let me know what you think! The Redoc look and feel seems a lot nicer to me than Swagger did.


api/rest/triple.yaml
```diff
@@ -0,0 +1,105 @@
+---
+swagger: '2.0'
```
beets, line 2: I find the yaml format very hard to follow along (and update in the future). Is there a guide for how you wrote this? I see that swagger has an editor - did you use that to craft this?

e.g. there doesn't seem to be a request sample for the GET call, but I'm not sure where that should go

beets: It's also important to keep the note about unraveling the payload - it's a big source of confusion for the API and what lead to us adding the fiddle examples.

beets (commented):
Thanks for this, and sorry for the delayed feedback. I'm wondering if there's more we can do to customize it so it works better on narrower screens, especially taking into account the main sidebar. Can we iterate on it more before updating content for more API calls? Happy to work together offline on this too.

KilimAnnejaro:
@beets, thanks for this feedback! I've moved the content I have into three new PRs on the api-update branch:

- https://github.com/datacommonsorg/docsite/pull/91
- https://github.com/datacommonsorg/docsite/pull/90
- https://github.com/datacommonsorg/docsite/pull/89

api/rest/query.html
```diff
@@ -0,0 +1,31 @@
+---
```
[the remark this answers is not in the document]
KilimAnnejaro: Yes, that is possible! Here's a PR: https://github.com/datacommonsorg/docsite/pull/93.

api/rest/query.html
```diff
@@ -0,0 +1,31 @@
...
+layout: default
+title: SPARQL
+nav_order: 1
+parent: REST
+grand_parent: API
+---
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>ReDoc</title>
```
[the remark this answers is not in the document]
KilimAnnejaro: I modified that in https://github.com/datacommonsorg/docsite/pull/93.

api/rest/triple.yaml
KilimAnnejaro: Hi Carolyn, these are great questions! I'll start with the YAML format one.

These YAML docs are hand-written following the OpenAPI format, which is just a standardized way of describing REST APIs. I personally thought the explanations at https://idratherbewriting.com/learnapidoc/restapispecifications.html and the more in-depth documentation at https://swagger.io/docs/specification/about/ were pretty helpful in understanding how to put a spec together.

I actually didn't use the Swagger tool to put these together, although I could have. Another option that I didn't experiment with was Stoplight: https://idratherbewriting.com/learnapidoc/pubapis_openapis_quickstart_stoplight.html.

KilimAnnejaro: For GET call request samples: I've added one each to #90 and #92. Let me know what you think!

KilimAnnejaro: For unraveling the payload: I'm not exactly sure what you mean by this--could you provide a clarifying example? Thank you!

beets: Thanks Anne - I'll take a look at the other PR's and the pointers!

I'd like these to be easily maintainable moving forward. The SPARQL and triples endpoint might have been the hardest to start with, so it does highlight some limits with this, namely:
- Formatting complex requests / responses: It's hard to read and understand the sample response for the triples query, especially since it's a long string blob. In our current doc, we're able to say: the payload is a long string blob, but if you unravel it, this is the json object you will get. Similarly, the example SPARQL request is a long string, where it is easier to read and understand it if it's multi-line
- Being able to add notes and other relevant info (maybe there is a way to do this).

We should also continue finding a more appropriate skin for this. The triple columns do not work for most screens (it only works on my huge monitor).

I'll add notes on the subsequent PR's, but would be great to keep the information that we have documented so far - such as "what's a triple", what the object model is in the json response, etc.

Sidenote, re: unraveling - see both the NOTE in the existing doc, and the sample code in the jsfiddle - there is two consecutive JSON.parse required.

Thanks!

KilimAnnejaro: Hi Carolyn, thanks for these comments! I've made some changes on the property value PR; I plan to focus on editing that one until we can get the formatting right.

Re unraveling, I went ahead and added the JS example as a code sample for the response in property value.

For skin, I'm honestly not sure what to do. Per https://github.com/Redocly/redoc/issues/1234, it seems there is no native way to remove the left column. I can try to remove with CSS modifications?

KilimAnnejaro:
Per our discussion yesterday, I'm going to go ahead and close the formatting PRs and focus on content, not formatting, going forward.

PR #90, opened 2020-11-07
property_value_updates

KilimAnnejaro:
Updating API docs for property value endpoint.


beets (commented):
Thanks Anne - please see comments from #92 which apply here as well. Should we try to iterate on one of these before spending time on other API endpoints?

KilimAnnejaro:
Closed per discussion (we can always reopen later).

PR #92, opened 2020-11-07
Updating API docs for the triple endpoint

KilimAnnejaro:


api/rest/triple.yaml
```diff
@@ -0,0 +1,109 @@
...
+  PostBody:
+    type: object
+    required:
+    - dcids
+    properties:
+      dcids:
+        type: array
+        items:
+          type: string
+        description: DCIDs targeted by the query.
```
beets, line 103: Could we link to a pointer of what is a DCID?

api/rest/triple.yaml
```diff
@@ -0,0 +1,109 @@
...
+      limit:
+        type: integer
+        description: Limit on number of triples desired in response.
+      key:
+        type: string
+        description: API key.
```
beets, line 109: this is an optional param - can we mark it as such, or add it to the description: "API key (optional)."

api/rest/triple.yaml
```diff
@@ -0,0 +1,109 @@
...
+        '200':
+          description: A string encoding of a JSON object keyed by the given DCIDs with the value a list of associated triples.
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  payload:
+                    type: string
+                    example: '{"dc/bsmvthtq89217":[{"subjectId":"dc/bsmvthtq89217","predicate":"country","objectId":"country/AUS","objectName":"Australia","objectTypes":["Country"],"provenanceId":"dc/93qydx3"},{"subjectId":"dc/bsmvthtq89217","predicate":"continent","objectId":"oceania","objectName":"Oceania","objectTypes":["Continent"],"provenanceId":"dc/93qydx3"},{"subjectId":"dc/bsmvthtq89217","predicate":"typeOf","objectId":"BiologicalSpecimen","objectName":"BiologicalSpecimen","objectTypes":["Class"],"provenanceId":"dc/93qydx3"},{"subjectId":"dc/bsmvthtq89217","predicate":"provenance","objectId":"dc/93qydx3","objectName":"http://nybg.org/","objectTypes":["Provenance"],"provenanceId":"dc/93qydx3"}]}'
```
beets, line 85: Can we format this better? Not sure if it's possible to add line breaks in here. Technically, the returned response is full of escape characters.

I wonder if we can do something similar to what we have, where we just say you get a string blob, but here's what it is if you json.parse it again.

api/rest/triple.yaml
```diff
@@ -0,0 +1,109 @@
+---
+swagger: '2.0'
+info:
+  title: Data Commons API
+  description: Given a list of nodes, return triples which are associated with the specified node(s).
```
beets, line 5: Can we add a short description of triples, or a pointer to one? Put another way, should this page standalone on it's own - if a user were to only see this page, should they know what to do without reading the rest of the site? Some questions might be: what is a triple? what is a dcid? Perhaps this is something we can address in a nicer intro page to the API.

api/rest/triple.yaml
```diff
@@ -0,0 +1,109 @@
...
+  version: 1.0.0
+host: api.datacommons.org
+schemes:
+- https
+paths:
+  "/node/triples":
+    get:
+      summary: Given a list of nodes, return triples which are associated with the
+        specified node(s).
+      description: Intended for users willing to modify the URLs called from their
```
beets, line 15: I'm not sure if we should focus all our REST descriptions with cURL - there are many ways to access REST APIs. Along those lines, that is one reason for using GET, though it is also useful for shorter queries. POST was added for longer queries that won't get bound by the URL length limit.

api/rest/triple.yaml
```diff
@@ -0,0 +1,109 @@
...
+        type: array
+        items:
+          type: string
+        required: true
+        description: DCIDs targeted by the query. Enter individually (for example, either `dcids=geoId/05&dcids=geoId/27` or `dcids=dc/bsmvthtq89217` would be acceptable query parameter syntax).
+      - in: query
+        name: limit
+        type: integer
+        required: false
+        description: Limit on number of triples desired in response.
```
beets, line 35: please don't forget to mention that the limit is 500 (see current doc)

api/rest/triple.yaml
```diff
@@ -0,0 +1,109 @@
...
+                properties:
+                  payload:
+                    type: string
+                    example: '{"dc/bsmvthtq89217":[{"subjectId":"dc/bsmvthtq89217","predicate":"country","objectId":"country/AUS","objectName":"Australia","objectTypes":["Country"],"provenanceId":"dc/93qydx3"},{"subjectId":"dc/bsmvthtq89217","predicate":"continent","objectId":"oceania","objectName":"Oceania","objectTypes":["Continent"],"provenanceId":"dc/93qydx3"},{"subjectId":"dc/bsmvthtq89217","predicate":"typeOf","objectId":"BiologicalSpecimen","objectName":"BiologicalSpecimen","objectTypes":["Class"],"provenanceId":"dc/93qydx3"},{"subjectId":"dc/bsmvthtq89217","predicate":"provenance","objectId":"dc/93qydx3","objectName":"http://nybg.org/","objectTypes":["Provenance"],"provenanceId":"dc/93qydx3"}]}'
+        '400':
+          description: Missing argument 'dcids'.    
+    post:
+      summary: Given a list of nodes, return triples which are associated with the
+        specified node(s).
+      description: Intended for users who would like to modify only the bodies of
```
beets, line 59: ditto earlier comment re: curl

beets (commented):
Thanks! 

It would be great if we could at least keep the information we have in our existing doc, especially the note on parsing the response, and the expected fields in the parsed JSON response. Would be nice if we could incorporate the JS fiddle example, which describes having to JSON.parse twice on the response. This has been a huge source of confusion with users of the API.

KilimAnnejaro:
Closed per discussion.

PR #93, opened 2020-11-07
Layout for REST docs

KilimAnnejaro:
Carolyn asked if there was a way we could convert the boilerplate HTML basis for the Redoc REST API docs into a layout file. This PR implements that idea.

To use, we will replace the html files with markup that looks like this example:

```
---
layout: restdocs
title: Triple
nav_order: 2
yaml_title: triple.yaml
parent: REST
grand_parent: API
---
```

### DIFF
--- a/_layouts/restdocs.html
+++ b/_layouts/restdocs.html
@@ -4,7 +4,7 @@ layout: default
 <!DOCTYPE html>
 <html>
   <head>
-    <title>Docs -- Data Commons</title>
+    <title>Docs - Data Commons</title>
     <!-- needed for adaptive design -->
     <meta charset="utf-8"/>
     <meta name="viewport" content="width=device-width, initial-scale=1">
--- a/_layouts/restdocs.html
+++ b/_layouts/restdocs.html
@@ -1,0 +1,27 @@
+---
+layout: default
+---
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Docs -- Data Commons</title>
+    <!-- needed for adaptive design -->
+    <meta charset="utf-8"/>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <link href="https://fonts.googleapis.com/css2?family=Public+Sans:wght@600&family=Roboto&display=swap" rel="stylesheet">
+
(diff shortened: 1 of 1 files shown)


_layouts/restdocs.html
```diff
@@ -0,0 +1,27 @@
+---
+layout: default
+---
+<!DOCTYPE html>
```
beets, line 4: This is added within the context of the default template, that is: we now have an embedded <html><head> etc within an HTML document. I believe all you need to add is the redoc and script tags.

_layouts/restdocs.html
```diff
@@ -0,0 +1,27 @@
...
+    <meta charset="utf-8"/>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <link href="https://fonts.googleapis.com/css2?family=Public+Sans:wght@600&family=Roboto&display=swap" rel="stylesheet">
+
+    <!--
+    ReDoc doesn't change outer page styles
+    -->
+    <style>
+      body {
+        margin: 0;
```
beets, line 18: I'm not sure if this is necessary (margin:0 is already added by bootstrap and removing padding:0 in the console doesn't seem to have a visible difference).

beets (commented):
Thanks - this is great. For the filename - it should be a singular noun (e.g. post, blog, etc). How about "redoc" since it's specific to that app?

beets:
Is it also possible to tweak the layout? Can we drop the inner left-hand nav column? The combined 4 column layout makes it very hard to read on even large monitors but it looks reasonable when the window size is under 800px wide (though with lots of extraneous vertical whitespace and issues with the table formatting).

What are your thoughts on how well this works with the API split across multiple pages? The "download API spec" link is not as effective, and the title of each page is Data Commons API 1.0.0 which is not useful. Authentication is also very prominent even though it's optional, and we don't get to clearly say "this is the GET vs POST" of each call without the nav menu.

It would also be nice to keep the styling somewhat consistent with our other APIs. Is there a way to control this?

KilimAnnejaro:
Closed per discussion.

PR #95, opened 2020-11-15
Polished first pass on property value REST docs

KilimAnnejaro:
First pass on content rewrite for the property value REST API docs. Comments:

- I was unsure of the explanatory verbiage for the direction field--feedback, suggestions?
- I formatted the JSON responses to look pretty, in the process changing some of the structure of the response. (For example, the strings returned are represented as JSON objects without backslash characters.) Let me know if you'd like me to change this.

Thank you!

---
This change is [Reviewable](https://reviewable.io/reviews/datacommonsorg/docsite/95)


api/rest/property_value.md
```diff
@@ -6,156 +6,492 @@ parent: REST
...
+        },
+        {
+          "dcid": "earthquake/usp000h6zw",
+          "name": "Madagascar",
+          "provenanceId": "dc/xz8ndk3",
+          "types": [
+            "EarthquakeEvent"
+          ]
+        },
+        {
```
beets, line 248: I'm not sure if it's illustrative to include the entire example here. Could we use ellipses like you did in the blocks above?

api/rest/property_value.md
```diff
@@ -6,156 +6,492 @@ parent: REST
...
-
 *   `key`: Your API key.
 
-## GET Request
-
-**Example**
+## How to construct a request to the property value endpoint
 
-*   Get name of two states
+### Step 1: assembling the information you will need
```
beets, line 36: Nit: Sentence casing "assembling the ...", also for other steps.

api/rest/property_value.md
```diff
@@ -6,156 +6,492 @@ parent: REST
...
+This endpoint requires two arguments and offers four additional optional arguments, as listed:
 
-*   Query with type filter and limit constraint
+ - `dcids`: Data Commons uniquely identifies nodes by assigning them DCIDs, or Data Commons IDs. Your query will need to specify the DCIDs for the nodes of interest.
+ - `property`: The property whose value you are interested in.
 
-    ```bash
-    curl 'https://api.datacommons.org/node/property-values?dcids=geoId/05&property=location&valueType=Election&limit=5'
-    ```
+In addition to these required properties, this endpoint also allows four optional arguments.
```
beets, line 43: Nit: maybe we shouldn't specify "four" if the number of arguments changes.

api/rest/property_value.md
```diff
@@ -6,156 +6,492 @@ parent: REST
...
+Your response will always look like this:
 
 ```json
 {
     "payload": "<payload string>",
 }
 ```
 
-The equivalent JSON of "payload string" is as follows.
+Here `"<payload string>"` is replaced by JSON, whose structure changes depending on whether the response contains node references.
```
beets, line 67: Rather than "replaced by JSON", the payload string is actually encoded JSON. I think it's important to mention that it needs to be decoded, and how to do so. While the jsfiddle had the example for JS, we could also add examples for other methods (e.g. cmd line, to go along with the cURL examples).

api/rest/property_value.md
```diff
@@ -6,156 +6,492 @@ parent: REST
...
-        "out": [
-          {
-            "value": "Arkansas",
-            "provenanceId": "dc/sm3m2w3"
-          }
-        ]
-    },
-    "geoId/06": {
-        "out": [
+    "dcid": {
```
beets, line 73: I think we need to be very clear that this is what you get _after_ decoding the payload.

api/rest/property_value.md
```diff
@@ -6,156 +6,492 @@ parent: REST
...
+            ...
         ]
+    },
+    ...
+}
+```
+
+**NOTES:** 
+ - The `provenanceId` is the DCID of the provenance for the corresponding value.
+ - You can run `JSON.parse()` on the `payload` field to retrieve the data. For example, in JavaScript: `var data = JSON.parse(response['payload'])`.
```
beets, line 108: Should we move up these notes, so it is clear how to get to the queried value?

api/rest/property_value.md
```diff
@@ -6,156 +6,492 @@ parent: REST
...
+  --header 'content-type: application/json' \
+  --data '{
+	"dcids": [
+		"dc/bsmvthtq89217"
+	],
+	"property": "order"
+}'
+```
+
+#### Response
```
beets, line 135: We should be very clear that this is after decoding, and that there is two steps involved, otherwise users will be very confused.

api/rest/property_value.md
```diff
@@ -6,156 +6,492 @@ parent: REST
...
-    ```
+In addition to these required properties, this endpoint also allows four optional arguments.
 
-*   Query with direction
+  - `valueType`: If the property queried only takes on node values, you can use this argument to filter nodes in the response, ensuring the response only contains nodes with the specified type.
 
-    ```bash
-    curl 'https://api.datacommons.org/node/property-values?dcids=geoId/05&property=containedInPlace&direction=out'
-    ```
+  - `direction`: You can specify this argument as `out` to indicate that you desire the response to only include nodes which are supercategories of the specified `DCIDs`, or `in` to only return nodes that are subcategories of the specified `DCIDs`. (For example, South America is a supercategory of Argentina, which in turn is a supercategory of Buenos Aires.)
```
beets, line 47: I think this is a good example. I wonder if it would be even clearer with a small node image, with pointed arrows to indicate direction, and some example queries (i.e. how does a user translate that example into property value queries?)

api/rest/property_value.md
```diff
@@ -6,156 +6,492 @@ parent: REST
...
-For each node, `in` contains nodes directed towards the node while `out`
-contains nodes/values directed away from the node. When the property value is a
-node, the result contains node information like `dcid`, `name`, `provenanceId`
-and `types`; when the property value is not a node, it has `value` and
-`provenanceId`.
+### Example 2: Retrieve the addresses of Stuyvesant High School in New York and Gunn High School in California.
+
+#### GET Request
+
+```curl
```
beets, line 156: Using "bash" instead of "curl" here would enable syntax highlighting

beets (commented):
Thanks Anne, this is really shaping up! I like the structure and flow of the page.

I commented inline about the 'direction' field (which was a source of confusion for me when I started).

Regarding the JSON responses - I think it's also a source of confusion and we shouldn't change the structure of the response so that users aren't thrown off when they try to replicate the examples. 

I also wonder if we can be clearer with examples, especially on how to get at the result you're looking for when you do a query? Especially for the simpler cases, where e.g. "get me the name of geoId/06" -- how do I go from response to result (California). We were asked quite explicitly for examples, which is why we added the JSFiddle, though that was only in JS, and we could add examples for other methods too (cmd-line, etc)

api/rest/property_value.md
tjann, line 67: We can consider adding/linking to documentation that talks about how there are only 2 values in DC: references and primitives.

tjann (commented):
Thanks for adding all the examples!

api/rest/property_value.md
```diff
@@ -6,156 +6,492 @@ parent: REST
 grand_parent: API
 ---
 
-# Show Property Values for Node(s)
+# Retrieve property values
 
-Given a list of nodes and a property label, returns values associated with the
-given property for each node.
+This endpoint is suitable for situations in which you have a node or list of nodes and desire to obtain the values of specified properties attached to those nodes.
```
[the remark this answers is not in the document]
beets: In REST parlance, everything is an endpoint, i.e. /node/property-values is an endpoint, and there isn't a concept of a "function" per se. I thought this had a good explanation: https://www.ibm.com/support/knowledgecenter/en/SSWLGF_8.5.0/com.ibm.sr.doc/rwsr_rest_model.html

api/rest/property_value.md
KilimAnnejaro: I added an image next to Example 5, the query that determines the country in which Buenos Aires is located. Does that location seem good to you? Furthermore, would you like for me to also add queries for the other three arrows on the diagram (ie, the continent on which Argentina is located, countries on the continent of South America, and also cities/administrative areas within Argentina)?

api/rest/property_value.md
```diff
@@ -6,156 +6,492 @@ parent: REST
...
 
-    ```bash
-    curl 'https://api.datacommons.org/node/property-values?dcids=geoId/05&property=containedInPlace&direction=out'
-    ```
+  - `direction`: You can specify this argument as `out` to indicate that you desire the response to only include nodes which are supercategories of the specified `DCIDs`, or `in` to only return nodes that are subcategories of the specified `DCIDs`. (For example, South America is a supercategory of Argentina, which in turn is a supercategory of Buenos Aires.)
+  
+  - `limit`: (≤ 500) Maximum number of values returned per node.
 
-## POST Request
+  - `key`: Your API key.
```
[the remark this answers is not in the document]
KilimAnnejaro: Yep, that's a great point! I've modified this section to remove a lot of the repetition. Let me know what you think of the restructuring!

api/rest/property_value.md
KilimAnnejaro: adjusted!

api/rest/property_value.md
KilimAnnejaro: I've added language to indicate the hard decoding requirement.

api/rest/property_value.md
KilimAnnejaro: Added that language.

api/rest/property_value.md
KilimAnnejaro: I merged these notes into the text.

beets (commented):
Thanks for adding the tabs, in general this is looking good!

> *[Reviewable](https://reviewable.io/reviews/datacommonsorg/docsite/95#-:-MMv_zS83o1_nAbqKOiL:bp9bg3b)* status: 0 of 8 files reviewed, 24 unresolved discussions (waiting on @beets, @KilimAnnejaro, @shifucun, and @tjann)

___
*[_layouts/default.html, line 11 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvR4fiB0rVKJ2F7V4X:-MMvR4fj3IZNQFtzaiRC:b7i3zfd) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/_layouts/default.html#L11)):*
> ```HTML
>   <link href="https://fonts.googleapis.com/css2?family=Public+Sans:wght@600&family=Roboto&display=swap" rel="stylesheet">
>   <link href="/assets/css/styles.css" rel="stylesheet">
>   <link rel="stylesheet" href="/assets/css/custom.css">
> ```

We shouldn't add additional css downloads. See comment in custom.css for a pointer to add it to the styles.css bundle
___
*[api/rest/property_value.md, line 43 at r1](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvQxweTTtgqffrHi9u-r1-43:-MMvU8-G3C4Y9kMCfUbk:bgxpgzh) ([raw file](https://github.com/datacommonsorg/docsite/blob/8c0b27533823a10f78fa59230102091e6fed4c38/api/rest/property_value.md#L43)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

Nit: maybe we shouldn't specify "four" if the number of arguments changes.
</blockquote></details>

Maybe: Here are helpful arguments ... (to avoid forgetting to update the count in the future when this list changes)
___
*[api/rest/property_value.md, line 47 at r1](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvQxweTTtgqffrHi9u-r1-47:-MMvUPbk4LPo-VXYcxxz:b393daq) ([raw file](https://github.com/datacommonsorg/docsite/blob/8c0b27533823a10f78fa59230102091e6fed4c38/api/rest/property_value.md#L47)):*
<details><summary><i>Previously, KilimAnnejaro (Anne E. Ulrich) wrote…</i></summary><blockquote>

I added an image next to Example 5, the query that determines the country in which Buenos Aires is located. Does that location seem good to you? Furthermore, would you like for me to also add queries for the other three arrows on the diagram (ie, the continent on which Argentina is located, countries on the continent of South America, and also cities/administrative areas within Argentina)?
</blockquote></details>

That looks great! Yes I think adding the property that corresponds to each edge will be helpful (and we should point to the example below that uses this diagram)
___
*[api/rest/property_value.md, line 73 at r1](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvQxweTTtgqffrHi9u-r1-73:-MMvV44p2GRDAx8-gKI0:b9y27z0) ([raw file](https://github.com/datacommonsorg/docsite/blob/8c0b27533823a10f78fa59230102091e6fed4c38/api/rest/property_value.md#L73)):*
<details><summary><i>Previously, KilimAnnejaro (Anne E. Ulrich) wrote…</i></summary><blockquote>

I've added language to indicate the hard decoding requirement.
</blockquote></details>

great, thanks!
___
*[api/rest/property_value.md, line 30 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvTlFY07y3s1qZpQhh:-MMvTlFY07y3s1qZpQhi:b-2ndu39) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/api/rest/property_value.md#L30)):*
> ```
> valued as
> ```
Maybe: Only valid values are ... or Only values allowed are ...
___
*[api/rest/property_value.md, line 73 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvV6ra-F6mBEfaGF7t:-MMvV6ra-F6mBEfaGF7u:bo4zlh) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/api/rest/property_value.md#L73)):*
> ```Markdown
> {
>     "dcid": {
>         "direction": [
> ```

We should try to differentiate between when it's the actual key vs when it's a string that contains a specified value, e.g. top level "dcid" is actually the dcid you queried for, vs "dcid" in the innermost object is the key "dcid". Maybe "<dcid>" and "dcid"? Thoughts?
___
*[api/rest/property_value.md, line 140 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvW1pV7Qm7-8gryuR0:-MMvW1pV7Qm7-8gryuR1:bqgn9pw) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/api/rest/property_value.md#L140)):*
> ```
> js
> ```

Should this be "JavaScript"?
___
*[api/rest/property_value.md, line 329 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvWx1w3aLZ2zlYgtHF:-MMvWx1w3aLZ2zlYgtHG:bjqhqc0) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/api/rest/property_value.md#L329)):*
> ```
> earthquakes
> ```
Nit: earthquake events
___
*[api/rest/property_value.md, line 404 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvX19R3BgPso3a-6pZ:-MMvX19R3BgPso3a-6p_:b-ls464) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/api/rest/property_value.md#L404)):*
> ```
> cyclone
> ```
Nit: cyclone event
___
*[assets/css/custom.css, line 1 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvRUbb5z5lQgWxYQ0T:-MMvRUbb5z5lQgWxYQ0U:b-lc3gd9) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/assets/css/custom.css#L1)):*
> ```CSS
> .tab {
> ```

Perhaps this file should be renamed to tabs.scss (and we can import it into styles.scss - jekyll will compile it into one css module for serving)
___
*[assets/images/SAmWithDir.png, line 0 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvSkij1dBFzjm2zwDB:-MMvSkij1dBFzjm2zwDC:b6scieq) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/assets/images/SAmWithDir.png#L0)):*
Can we add a new subdirectory for this image (assets/images/rest) and follow the existing naming scheme?
e.g. property_value_direction_example.png

I wonder if it would be helpful to also add the property corresponding to each edge.
___
*[assets/js/tabs.js, line 1 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvXHavEWEqZ0aP0ctp:-MMvXHavEWEqZ0aP0ctq:blz4k6w) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/assets/js/tabs.js#L1)):*
> ```JavaScript
> const removeActiveClasses = function (ulElement) {
> ```

Please add a top-level comment describing the contents of this module and each function.
___
*[assets/js/tabs.js, line 6 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMv_aTwCp7lhUYC5oFv:-MMv_aTwCp7lhUYC5oFw:bmpksmo) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/assets/js/tabs.js#L6)):*
> ```JavaScript
>         li.classList.remove('active');
>     });
>   }
> ```

Please watch the indentation for here and the following function
___
*[assets/js/tabs.js, line 10 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvYfGQAf7t3hK0ibH5:-MMvYfGQAf7t3hK0ibH6:b-52xdsf) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/assets/js/tabs.js#L10)):*
> ```JavaScript
>   const getChildPosition = function (element) {
>         var parent = element.parentNode;
>         var i = 0;
> ```

This line isn't necessary since i is redefined in the for loop
___
*[assets/js/tabs.js, line 16 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvYk7Q-2LaVEzVXz7P:-MMvYk7RAQxDzB7HJsl_:b-zc710m) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/assets/js/tabs.js#L16)):*
> ```JavaScript
>             }
>         }
> ```

Maybe add a comment that this should not be reachable?
___
*[assets/js/tabs.js, line 21 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvZksiEPRjhIVquFwX:-MMvZksiEPRjhIVquFwY:blb857c) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/assets/js/tabs.js#L21)):*
> ```JavaScript
> 
> window.addEventListener('load', function () {
>     const tabLinks = document.querySelectorAll('ul.tab li a');
> ```

I think the code would be simpler if we installed the onclick on the li element.
___
*[assets/js/tabs.js, line 30 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvYolV0zjLU1fk8wsT:-MMvYolV0zjLU1fk8wsU:b-ghndxh) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/assets/js/tabs.js#L30)):*
> ```JavaScript
>         ulTab = liTab.parentNode;
>         position = getChildPosition(liTab);
>         if (liTab.className.includes('active')) {
> ```

Maybe move this two lines up (after liTab). Since ulTab and position isn't necessary if we early return.
___
*[assets/js/tabs.js, line 39 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMv_JOu1losvOVM9u9i:-MMv_JOu1losvOVM9u9j:b-3cway3) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/assets/js/tabs.js#L39)):*
> ```JavaScript
>         removeActiveClasses(tabContentElement);
> 
>         tabContentElement.querySelectorAll('li')[position].classList.add('active');
> ```

Perhaps I'm missing something, but is this the equivalent of: `liTab.classList.add('active')` (we might not need position after all)

beets:
I apologize, I should've mentioned that I'm testing out using reviewable.io for the PR review. I thought it would have synced the comments inline with the file viewer here. We can keep things in github PR if you prefer, let me know!

KilimAnnejaro (commented):
> *[Reviewable](https://reviewable.io/reviews/datacommonsorg/docsite/95)* status: 0 of 8 files reviewed, 20 unresolved discussions (waiting on @beets, @KilimAnnejaro, @shifucun, and @tjann)

___
*[api/rest/property_value.md, line 30 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvTlFY07y3s1qZpQhh:-MMxN2Vu21h-p0nM6Mu5:b-896fix) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/api/rest/property_value.md#L30)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

> ```
> valued as
> ```
Maybe: Only valid values are ... or Only values allowed are ...
</blockquote></details>

Done.
___
*[api/rest/property_value.md, line 73 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvV6ra-F6mBEfaGF7t:-MMxWr28CDL7erx_2d6n:b-wou0z) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/api/rest/property_value.md#L73)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

We should try to differentiate between when it's the actual key vs when it's a string that contains a specified value, e.g. top level "dcid" is actually the dcid you queried for, vs "dcid" in the innermost object is the key "dcid". Maybe "<dcid>" and "dcid"? Thoughts?
</blockquote></details>

Yep, this is a great point! I've changed dcid here to <dcid>, which maybe isn't perfect since it implied dcid is a type--but it seems logical from the <payload string> line above? Let me know what you think!
___
*[api/rest/property_value.md, line 140 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvW1pV7Qm7-8gryuR0:-MMxXbPk2vN7Cq-Y0w_s:b-d2w7qu) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/api/rest/property_value.md#L140)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

> ```
> js
> ```

Should this be "JavaScript"?
</blockquote></details>

So the examples at https://github.com/Ovski4/jekyll-tabs used JS for the tab name; if you think JavaScript would make more sense, we can absolutely do that!
___
*[assets/images/SAmWithDir.png, line  at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvSkij1dBFzjm2zwDB:-MMxaXraA9uN-thLJSNl:b-yuy14x) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/assets/images/SAmWithDir.png#L)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

Can we add a new subdirectory for this image (assets/images/rest) and follow the existing naming scheme?
e.g. property_value_direction_example.png

I wonder if it would be helpful to also add the property corresponding to each edge.
</blockquote></details>

I went ahead and created the subdirectory and renamed the file.

Going off this comment and the one earlier, I guess I am a little confused about how to add the property corresponding to each edge. Each edge is labelled with a direction, so I suppose I could replace the "in" direction labels with "containedInPlace" to indicate that the place pointed to has the property "containedInPlace" with respect to the place pointed from. However, I'm not sure how to indicate that for the "out" direction, since I am not able to find any "containedOutPlace" property or similar. What I could do (and indeed did) is add content to the caption indicating these thoughts. However, now the caption seems very long and ugly to me. Do you have any thoughts on this?

KilimAnnejaro (commented):
> *[Reviewable](https://reviewable.io/reviews/datacommonsorg/docsite/95)* status: 0 of 8 files reviewed, 20 unresolved discussions (waiting on @beets, @KilimAnnejaro, @shifucun, and @tjann)

___
*[assets/js/tabs.js, line 1 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvXHavEWEqZ0aP0ctp:-MMxyMzOANwzsuKsOPoG:b-896fix) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/assets/js/tabs.js#L1)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

Please add a top-level comment describing the contents of this module and each function.
</blockquote></details>

Done.
___
*[assets/js/tabs.js, line 6 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMv_aTwCp7lhUYC5oFv:-MMxnUBm33wLE8tkToBS:b-896fix) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/assets/js/tabs.js#L6)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

Please watch the indentation for here and the following function
</blockquote></details>

Done.
___
*[assets/js/tabs.js, line 10 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvYfGQAf7t3hK0ibH5:-MMxnTMe1RwUgGVrLA0q:b-896fix) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/assets/js/tabs.js#L10)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

This line isn't necessary since i is redefined in the for loop
</blockquote></details>

Done.
___
*[assets/js/tabs.js, line 16 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvYk7Q-2LaVEzVXz7P:-MMxo7ak9banWWyISajg:b-896fix) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/assets/js/tabs.js#L16)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

Maybe add a comment that this should not be reachable?
</blockquote></details>

Done.
___
*[assets/js/tabs.js, line 21 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvZksiEPRjhIVquFwX:-MMy0rTG3S98V3hpRdCA:b-fcbmu7) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/assets/js/tabs.js#L21)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

I think the code would be simpler if we installed the onclick on the li element.
</blockquote></details>

Could you clarify this? I am not sure what that would look like.
___
*[assets/js/tabs.js, line 30 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvYolV0zjLU1fk8wsT:-MMxur422cnI2A-WKzpe:b-896fix) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/assets/js/tabs.js#L30)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

Maybe move this two lines up (after liTab). Since ulTab and position isn't necessary if we early return.
</blockquote></details>

Done.
___
*[assets/js/tabs.js, line 39 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMv_JOu1losvOVM9u9i:-MMy-aaG4czdD_N0Ms2n:b-ui8fq9) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/assets/js/tabs.js#L39)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

Perhaps I'm missing something, but is this the equivalent of: `liTab.classList.add('active')` (we might not need position after all)
</blockquote></details>

So I tried commenting this line out as an experiment, and it appears that when you click on another tab, the content from the previous tab does not appear, but the old content is not replaced with the new. 
So I think the position is still needed because it ensures that the tabs and content are synced, but I am not sure--open to feedback as always though!

KilimAnnejaro (commented):
> *[Reviewable](https://reviewable.io/reviews/datacommonsorg/docsite/95)* status: 0 of 8 files reviewed, 20 unresolved discussions (waiting on @beets, @KilimAnnejaro, @shifucun, and @tjann)

___
*[assets/css/custom.css, line 1 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvRUbb5z5lQgWxYQ0T:-MMy2j1E38WmiIIhuJoI:b-wdlkjk) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/assets/css/custom.css#L1)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

Perhaps this file should be renamed to tabs.scss (and we can import it into styles.scss - jekyll will compile it into one css module for serving)
</blockquote></details>

I ended up directly merging the styles into the main CSS file.

KilimAnnejaro:
Hi @beets, I responded to all your comments within the Reviewable format. If you'd like to stick with Reviewable going forward, no worries; I personally like the Github comment/PR experience better, but I may just not know enough about Reviewable's features yet :wink:

KilimAnnejaro (commented):
> *[Reviewable](https://reviewable.io/reviews/datacommonsorg/docsite/95)* status: 0 of 8 files reviewed, 20 unresolved discussions (waiting on @beets, @KilimAnnejaro, @shifucun, and @tjann)

___
*[_layouts/default.html, line 11 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvR4fiB0rVKJ2F7V4X:-MMy4-dMBAhWEqS1UF3w:b-896fix) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/_layouts/default.html#L11)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

We shouldn't add additional css downloads. See comment in custom.css for a pointer to add it to the styles.css bundle
</blockquote></details>

Done.

tjann (changes requested):
> Reviewed 3 of 7 files at r5, 1 of 6 files at r6.
> *[Reviewable](https://reviewable.io/reviews/datacommonsorg/docsite/95)* status: 3 of 7 files reviewed, 19 unresolved discussions (waiting on @beets, @KilimAnnejaro, and @tjann)

___
*[api/rest/property_value.md, line 73 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvV6ra-F6mBEfaGF7t:-MN-jy8a7ByOne1iGFZ7:bqvzqis) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/api/rest/property_value.md#L73)):*
<details><summary><i>Previously, KilimAnnejaro (Anne E. Ulrich) wrote…</i></summary><blockquote>

Yep, this is a great point! I've changed dcid here to <dcid>, which maybe isn't perfect since it implied dcid is a type--but it seems logical from the <payload string> line above? Let me know what you think!
</blockquote></details>

What if we kept it as an example decoded response instead of saying what DC type the strings will be? They could be foo bar.

like `MarcoPolo` for the <dcid> and `FooBarEncyclopedia` for provenance Id. @beets wdyt? I'm suggesting a large change so don't want Anne to do it and then we disagree :P  
___
*[api/rest/property_value.md, line 38 at r6](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MN-hbwy902pjMfzTx5Q:-MN-gpvr9rEbusISJTSm:byybk57) ([raw file](https://github.com/datacommonsorg/docsite/blob/3b643de3775204249a501518116a5e022d495428/api/rest/property_value.md#L38)):*
> ```Markdown
> ### Step 1: Assembling the information you will need
> 
> This endpoint requires two arguments and offers four additional optional arguments, as listed:
> ```

"Going into more detail about how to assemble the values for the required arguments:" 
___
*[api/rest/property_value.md, line 42 at r6](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MN-hrdAAl3dpEG4NHBi:-MN-hrdAAl3dpEG4NHBj:be5k1zq) ([raw file](https://github.com/datacommonsorg/docsite/blob/3b643de3775204249a501518116a5e022d495428/api/rest/property_value.md#L42)):*
> ```Markdown
>  - `dcids`: Data Commons uniquely identifies nodes by assigning them DCIDs, or Data Commons IDs. Your query will need to specify the DCIDs for the nodes of interest.
> 
>  - `property`: The property whose value you are interested in.
> ```

interested in, such as "name" for the name of a node, or "typeOf" for the type of a node.

KilimAnnejaro (commented):
> *[Reviewable](https://reviewable.io/reviews/datacommonsorg/docsite/95)* status: 3 of 7 files reviewed, 19 unresolved discussions (waiting on @beets, @KilimAnnejaro, and @tjann)

___
*[api/rest/property_value.md, line 38 at r6](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MN-hbwy902pjMfzTx5Q:-MN-rbmB9z4d1pM-YMYi:bsg6ln3) ([raw file](https://github.com/datacommonsorg/docsite/blob/3b643de3775204249a501518116a5e022d495428/api/rest/property_value.md#L38)):*
<details><summary><i>Previously, tjann (TJann) wrote…</i></summary><blockquote>

"Going into more detail about how to assemble the values for the required arguments:" 
</blockquote></details>

Whoops, my bad--I thought I'd cleaned this language up! I've updated as suggested.
___
*[api/rest/property_value.md, line 42 at r6](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MN-hrdAAl3dpEG4NHBi:-MN-s4vn5cezip7kXXBZ:b-896fix) ([raw file](https://github.com/datacommonsorg/docsite/blob/3b643de3775204249a501518116a5e022d495428/api/rest/property_value.md#L42)):*
<details><summary><i>Previously, tjann (TJann) wrote…</i></summary><blockquote>

interested in, such as "name" for the name of a node, or "typeOf" for the type of a node.
</blockquote></details>

Done.

api/rest/property_value.md
```diff
@@ -6,156 +6,568 @@ parent: REST
...
 
-**Auth required**: Optional
+**Authentication**: Optional
 
-**Required Arguments**:
+**Required arguments**:
 
 *   `dcids`: A list of nodes to query, identified by their DCID.
-
 *   `property`: The property label to query for.
```
tjann, line 24: I might suggest removing "label"

api/rest/property_value.md
```diff
@@ -6,156 +6,568 @@ parent: REST
...
+    },
+    ...
+}
+```
+
+**NOTE:** The `provenanceId` is the DCID of the provenance for the corresponding value.
+
+## Example requests and responses
+
+### Example 1: Retrieve the two common names of the country of Côte d'Ivoire.
```
tjann, line 109: Retrieve the common name(s) ? I don't want users to think we are doing limit=2

tjann (approved):
I think the content looks great! Approval for the content, but will keep monitoring to get remaining discussions closed.

beets (changes requested):
> *[Reviewable](https://reviewable.io/reviews/datacommonsorg/docsite/95)* status: 3 of 7 files reviewed, 10 unresolved discussions (waiting on @beets, @KilimAnnejaro, and @tjann)

___
*[api/rest/property_value.md, line 47 at r1](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvQxweTTtgqffrHi9u-r1-47:-MN0aTOuDhlW3xx3TZ_f:b-3m2ol8) ([raw file](https://github.com/datacommonsorg/docsite/blob/8c0b27533823a10f78fa59230102091e6fed4c38/api/rest/property_value.md#L47)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

That looks great! Yes I think adding the property that corresponds to each edge will be helpful (and we should point to the example below that uses this diagram)
</blockquote></details>

Hm, after looking at this more, can we use it in terms of containedInPlace? i.e. for the property `containedInPlace` on the node "Argentina", `in` results in Bueno Aires, `out` results in South America. And we could simplify that figure with just those two edges, with the containedInPlace label (since in/out is denoted by the arrow). 
___
*[api/rest/property_value.md, line 73 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvV6ra-F6mBEfaGF7t:-MN0XFUt8ap7qGLBirhU:b-tqu5ac) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/api/rest/property_value.md#L73)):*
<details><summary><i>Previously, tjann (TJann) wrote…</i></summary><blockquote>

What if we kept it as an example decoded response instead of saying what DC type the strings will be? They could be foo bar.

like `MarcoPolo` for the <dcid> and `FooBarEncyclopedia` for provenance Id. @beets wdyt? I'm suggesting a large change so don't want Anne to do it and then we disagree :P  
</blockquote></details>

I think <dcid> works! Also note that it should be <direction>. We can point to the example responses below.
___
*[api/rest/property_value.md, line 140 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvW1pV7Qm7-8gryuR0:-MN0X_2gAUCgAVYiyQoq:b8rqows) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/api/rest/property_value.md#L140)):*
<details><summary><i>Previously, KilimAnnejaro (Anne E. Ulrich) wrote…</i></summary><blockquote>

So the examples at https://github.com/Ovski4/jekyll-tabs used JS for the tab name; if you think JavaScript would make more sense, we can absolutely do that!
</blockquote></details>

I think JavaScript is more accurate...
___
*[api/rest/property_value.md, line 38 at r6](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MN-hbwy902pjMfzTx5Q:-MN0WVCj-zHT98qDpFAk:b-v381ya) ([raw file](https://github.com/datacommonsorg/docsite/blob/3b643de3775204249a501518116a5e022d495428/api/rest/property_value.md#L38)):*
<details><summary><i>Previously, KilimAnnejaro (Anne E. Ulrich) wrote…</i></summary><blockquote>

Whoops, my bad--I thought I'd cleaned this language up! I've updated as suggested.
</blockquote></details>

We should also try avoiding counting the arguments, though keeping "requires two arguments" seems reasonable.
___
*[api/rest/property_value.md, line 42 at r6](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MN-hrdAAl3dpEG4NHBi:-MN0Wq_v8kBgKHASnkt1:b-ruk13j) ([raw file](https://github.com/datacommonsorg/docsite/blob/3b643de3775204249a501518116a5e022d495428/api/rest/property_value.md#L42)):*
<details><summary><i>Previously, KilimAnnejaro (Anne E. Ulrich) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

I think @tjann was suggesting adding some possible properties in the description.
___
*[assets/js/tabs.js, line 1 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvXHavEWEqZ0aP0ctp:-MN0cEjc0joZGo0Hno0k:b-laf9gr) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/assets/js/tabs.js#L1)):*
<details><summary><i>Previously, KilimAnnejaro (Anne E. Ulrich) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Thanks! Could you move the docs for the functions to right above each function?
___
*[assets/js/tabs.js, line 21 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvZksiEPRjhIVquFwX:-MN0cdp11fQ9G_B35a7D:b-8y7weu) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/assets/js/tabs.js#L21)):*
<details><summary><i>Previously, KilimAnnejaro (Anne E. Ulrich) wrote…</i></summary><blockquote>

Could you clarify this? I am not sure what that would look like.
</blockquote></details>

It's alright, what we have here is good.
___
*[assets/js/tabs.js, line 39 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMv_JOu1losvOVM9u9i:-MN0cRn101-TZUMsiqm1:b-wjccdi) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/assets/js/tabs.js#L39)):*
<details><summary><i>Previously, KilimAnnejaro (Anne E. Ulrich) wrote…</i></summary><blockquote>

So I tried commenting this line out as an experiment, and it appears that when you click on another tab, the content from the previous tab does not appear, but the old content is not replaced with the new. So we have the `liTab.classList.add('active')` below to mark particular tabs as active and this line to mark particular content as active.

To ensure that the right tab is paired with the right content, we use the helper method to define this `position`, then activate both the content and the tab at that position. Based on this analysis, I guess I am not sure why position would be unnecessary--open to feedback as always though!
</blockquote></details>

Ah, yes! Sorry I missed that it is using the position on another ul/li tree. Ignore this comment then.
___
*[assets/images/SAmWithDir.png, line  at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvSkij1dBFzjm2zwDB:-MN0bwSuArfWLxFFFmTz:bkk32x8) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/assets/images/SAmWithDir.png#L)):*
<details><summary><i>Previously, KilimAnnejaro (Anne E. Ulrich) wrote…</i></summary><blockquote>

I went ahead and created the subdirectory and renamed the file.

Going off this comment and the one earlier, I guess I am a little confused about how to add the property corresponding to each edge. Each edge is labelled with a direction, so I suppose I could replace the "in" direction labels with "containedInPlace" to indicate that the place pointed to has the property "containedInPlace" with respect to the place pointed from. However, I'm not sure how to indicate that for the "out" direction, since I am not able to find any "containedOutPlace" property or similar. What I could do (and indeed did) is add content to the caption indicating these thoughts. However, now the caption seems very long and ugly to me. Do you have any thoughts on this?
</blockquote></details>

Added some suggestions to the comment above (where we were discussing the direction argument)

beets:
Could we try reviewable for this PR at least? We were finding that we missed out on unresolved comments when the PR's get large, and this at least highlights which are left unresolved.

beets:
Agree though that the content looks great! Just a few more things to resolve. Thanks for working through this with us Anne!

api/rest/property_value.md
KilimAnnejaro: I will remove "two" from this header.

KilimAnnejaro (commented):
Thank you!

> *[Reviewable](https://reviewable.io/reviews/datacommonsorg/docsite/95#-:-MN1Q0kS1nxjKLsfhl8k:b-9755fn)* status: 3 of 7 files reviewed, 10 unresolved discussions (waiting on @beets, @KilimAnnejaro, and @tjann)

___
*[api/rest/property_value.md, line 47 at r1](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvQxweTTtgqffrHi9u-r1-47:-MN1bzpeARV6SXdWM2w9:bl2lymm) ([raw file](https://github.com/datacommonsorg/docsite/blob/8c0b27533823a10f78fa59230102091e6fed4c38/api/rest/property_value.md#L47)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

Hm, after looking at this more, can we use it in terms of containedInPlace? i.e. for the property `containedInPlace` on the node "Argentina", `in` results in Bueno Aires, `out` results in South America. And we could simplify that figure with just those two edges, with the containedInPlace label (since in/out is denoted by the arrow). 
</blockquote></details>

I am not sure I understand what I should do about the diagram? But I did update this description to use the `containedInPlace` language rather than the supercategory/subcategory bits.
___
*[api/rest/property_value.md, line 73 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvV6ra-F6mBEfaGF7t:-MN1WUg3BbBUHrmD0NH4:b-oh3ezv) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/api/rest/property_value.md#L73)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

I think <dcid> works! Also note that it should be <direction>. We can point to the example responses below.
</blockquote></details>

I don't understand?
___
*[api/rest/property_value.md, line 140 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvW1pV7Qm7-8gryuR0:-MN1Ht9t-rgzZ8t5d1iR:b-9nw43w) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/api/rest/property_value.md#L140)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

I think JavaScript is more accurate...
</blockquote></details>

Yep, we can do that!
___
*[api/rest/property_value.md, line 38 at r6](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MN-hbwy902pjMfzTx5Q:-MN1QJrvE5y4DqPZeOL1:b-896fix) ([raw file](https://github.com/datacommonsorg/docsite/blob/3b643de3775204249a501518116a5e022d495428/api/rest/property_value.md#L38)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

We should also try avoiding counting the arguments, though keeping "requires two arguments" seems reasonable.
</blockquote></details>

Done.
___
*[api/rest/property_value.md, line 42 at r6](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MN-hrdAAl3dpEG4NHBi:-MN1QUc4EH5X4Hsrutc5:b-m3pcbe) ([raw file](https://github.com/datacommonsorg/docsite/blob/3b643de3775204249a501518116a5e022d495428/api/rest/property_value.md#L42)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

I think @tjann was suggesting adding some possible properties in the description.
</blockquote></details>

So I just checked on Github and it appears to me that I incorporated @tjann 's comment verbatim into the PR. Am I looking at the wrong page?
___
*[api/rest/property_value.md, line 24 at r7](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvQxweTTtgqffrHi9u-r7-24:-MN1Q56Z9cGlt4apPKA8:b-896fix) ([raw file](https://github.com/datacommonsorg/docsite/blob/eccf14d05a4f81b20d0eed558543d05c991f2d42/api/rest/property_value.md#L24)):*
<details><summary><i>Previously, tjann (TJann) wrote…</i></summary><blockquote>

I might suggest removing "label"
</blockquote></details>

Done.
___
*[api/rest/property_value.md, line 109 at r7](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvQxweTTtgqffrHi9u-r7-109:-MN1HQTv8yTbTxuGlkUC:b-896fix) ([raw file](https://github.com/datacommonsorg/docsite/blob/eccf14d05a4f81b20d0eed558543d05c991f2d42/api/rest/property_value.md#L109)):*
<details><summary><i>Previously, tjann (TJann) wrote…</i></summary><blockquote>

Retrieve the common name(s) ? I don't want users to think we are doing limit=2
</blockquote></details>

Done.
___
*[assets/js/tabs.js, line 1 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvXHavEWEqZ0aP0ctp:-MN1Punk4OFDKN5r9qxE:ba479fx) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/assets/js/tabs.js#L1)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

Thanks! Could you move the docs for the functions to right above each function?
</blockquote></details>

Yep, this is a good idea!

beets (changes requested):
Sorry for the late response on the PR!

> *[Reviewable](https://reviewable.io/reviews/datacommonsorg/docsite/95#-:-MNVJLatCNTvH9Yzrcsy:bdkqfk2)* status: 3 of 7 files reviewed, 10 unresolved discussions (waiting on @beets, @KilimAnnejaro, and @tjann)

___
*[api/rest/property_value.md, line 47 at r1](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvQxweTTtgqffrHi9u-r1-47:-MNVCmDj9SNxg2Gk8Pwk:b-yr6r65) ([raw file](https://github.com/datacommonsorg/docsite/blob/8c0b27533823a10f78fa59230102091e6fed4c38/api/rest/property_value.md#L47)):*
<details><summary><i>Previously, KilimAnnejaro (Anne E. Ulrich) wrote…</i></summary><blockquote>

I am not sure I understand what I should do about the diagram? But I did update this description to use the `containedInPlace` language rather than the supercategory/subcategory bits.
</blockquote></details>

How about something like this:

`direction`: This refers to the orientation, or direction, of the edge. You can specify this argument as `in` or `out`. (For example, the figure below illustrates  the directions for the property `containedInPlace` of the node for Argentina).

[ South America ] <-- containedInPlace -- [ Argentina ] <-- containedInPlace -- [ Buenos Aires ]

PropertyValue(containedInPlace, Argentina, 'in') --> Buenos Aires
PropertyValue(containedInPlace, Argneinta, 'out') --> South America 

(not sure if that pseudocode adds to the confusion.. but something to that effect?)
___
*[api/rest/property_value.md, line 73 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvV6ra-F6mBEfaGF7t:-MNVI0C8-_uNSZGCt5JC:bxlj5ow) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/api/rest/property_value.md#L73)):*
<details><summary><i>Previously, KilimAnnejaro (Anne E. Ulrich) wrote…</i></summary><blockquote>

I don't understand?
</blockquote></details>

Huh, maybe it was lost with the formatting. I meant to say, for direction -- it should also have the angle brackets since it will have the values "in" or "out".
___
*[api/rest/property_value.md, line 140 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvW1pV7Qm7-8gryuR0:-MNVIJ817JiSDmx-N9UZ:b-i11fkc) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/api/rest/property_value.md#L140)):*
<details><summary><i>Previously, KilimAnnejaro (Anne E. Ulrich) wrote…</i></summary><blockquote>

Yep, we can do that!
</blockquote></details>

Update to "JavaScript" (capitalized J and S)?
___
*[api/rest/property_value.md, line 42 at r6](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MN-hrdAAl3dpEG4NHBi:-MNVBxXi3d3G7KH1FhX0:b-bcojwo) ([raw file](https://github.com/datacommonsorg/docsite/blob/3b643de3775204249a501518116a5e022d495428/api/rest/property_value.md#L42)):*
<details><summary><i>Previously, KilimAnnejaro (Anne E. Ulrich) wrote…</i></summary><blockquote>

So I just checked on Github and it appears to me that I incorporated @tjann 's comment verbatim into the PR. Am I looking at the wrong page?
</blockquote></details>

Ah looks like it was added in the next commit. Thanks
___
*[api/rest/property_value.md, line 479 at r9](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MNVIWV55B7r8d2rFG2X:-MNVIWV55B7r8d2rFG2Y:b-4eyatl) ([raw file](https://github.com/datacommonsorg/docsite/blob/c26a8475c44a4bd7a8810b13ee23c2929cd6eb15/api/rest/property_value.md#L479)):*
> ```Markdown
> ```
> 
> ### Example 6: Retrieve the country in which Buenos Aires is located.
> ```

Took a read over of the examples again and I think we should update this to match the figure (which we should include here and above with the "direction" explanation)

Can we do this as the query for this example:
https://api.datacommons.org/node/property-values?dcids=country/ARG&property=containedInPlace

Note how there are results for both 'in' and 'out, which I think highlights the directionality well.
___
*[assets/images/SAmWithDir.png, line  at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvSkij1dBFzjm2zwDB:-MNVJ2TJ4MXfCH3Vc38t:b-3itoyf) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/assets/images/SAmWithDir.png#L)):*
Could this be a good diagram?

> [ South America ] <-- containedInPlace -- [ Argentina ] <-- containedInPlace -- [ Buenos Aires ]

KilimAnnejaro (commented):
no worries :smiley: 

> *[Reviewable](https://reviewable.io/reviews/datacommonsorg/docsite/95#-:-MNWWf-V3tnuKtmZwQMu:b-hfc6ga)* status: 3 of 7 files reviewed, 6 unresolved discussions (waiting on @beets and @KilimAnnejaro)

___
*[api/rest/property_value.md, line 47 at r1](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvQxweTTtgqffrHi9u-r1-47:-MNWd04dDTiVPseU-ff_:b-hx4e02) ([raw file](https://github.com/datacommonsorg/docsite/blob/8c0b27533823a10f78fa59230102091e6fed4c38/api/rest/property_value.md#L47)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

How about something like this:

`direction`: This refers to the orientation, or direction, of the edge. You can specify this argument as `in` or `out`. (For example, the figure below illustrates  the directions for the property `containedInPlace` of the node for Argentina).

[ South America ] <-- containedInPlace -- [ Argentina ] <-- containedInPlace -- [ Buenos Aires ]

PropertyValue(containedInPlace, Argentina, 'in') --> Buenos Aires
PropertyValue(containedInPlace, Argneinta, 'out') --> South America 

(not sure if that pseudocode adds to the confusion.. but something to that effect?)
</blockquote></details>

So I went ahead and English-ized the pseudocode into the caption and also modified the image as we discussed. I don't love the way the caption is worded, but I think it hits the main ideas.
___
*[api/rest/property_value.md, line 73 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvV6ra-F6mBEfaGF7t:-MNWdPyz5MRnsTcGin7m:b-896fix) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/api/rest/property_value.md#L73)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

Huh, maybe it was lost with the formatting. I meant to say, for direction -- it should also have the angle brackets since it will have the values "in" or "out".
</blockquote></details>

Done.
___
*[api/rest/property_value.md, line 140 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvW1pV7Qm7-8gryuR0:-MNVi4K9CtxYqB6kEXm6:b-896fix) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/api/rest/property_value.md#L140)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

Update to "JavaScript" (capitalized J and S)?
</blockquote></details>

Done.
___
*[assets/images/SAmWithDir.png, line  at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvSkij1dBFzjm2zwDB:-MNWdTP5AVZEa5_69XIi:b-k4o5mh) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/assets/images/SAmWithDir.png#L)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

Could this be a good diagram?

> [ South America ] <-- containedInPlace -- [ Argentina ] <-- containedInPlace -- [ Buenos Aires ]

</blockquote></details>

Changed the diagram and incorporated the suggestions!

KilimAnnejaro (commented):
> *[Reviewable](https://reviewable.io/reviews/datacommonsorg/docsite/95)* status: 3 of 7 files reviewed, 6 unresolved discussions (waiting on @beets)

___
*[api/rest/property_value.md, line 479 at r9](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MNVIWV55B7r8d2rFG2X:-MNWfTdDAaQoXD4Q8iGj:bpr880f) ([raw file](https://github.com/datacommonsorg/docsite/blob/c26a8475c44a4bd7a8810b13ee23c2929cd6eb15/api/rest/property_value.md#L479)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

Took a read over of the examples again and I think we should update this to match the figure (which we should include here and above with the "direction" explanation)

Can we do this as the query for this example:
https://api.datacommons.org/node/property-values?dcids=country/ARG&property=containedInPlace

Note how there are results for both 'in' and 'out, which I think highlights the directionality well.
</blockquote></details>

I've updated the query in the most recent commit.
I am a little concerned about repeating the figure here as well as at the top, since the page seems too short for such a repetition. However, I'm not sure--thoughts?

beets (changes requested):
> *[Reviewable](https://reviewable.io/reviews/datacommonsorg/docsite/95)* status: 3 of 7 files reviewed, 4 unresolved discussions (waiting on @beets, @KilimAnnejaro, and @tjann)

___
*[api/rest/property_value.md, line 47 at r1](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvQxweTTtgqffrHi9u-r1-47:-MNZk36G8RYmerf9EQDL:b-3w4doe) ([raw file](https://github.com/datacommonsorg/docsite/blob/8c0b27533823a10f78fa59230102091e6fed4c38/api/rest/property_value.md#L47)):*
<details><summary><i>Previously, KilimAnnejaro (Anne E. Ulrich) wrote…</i></summary><blockquote>

So I went ahead and English-ized the pseudocode into the caption and also modified the image as we discussed. I don't love the way the caption is worded, but I think it hits the main ideas.
</blockquote></details>

With the suggestion in my last comment, I also wanted to clarify that this doesn't only apply to "containedInPlace" -- this applies to all properties (the second sentence of the description). Could you update that as well?
___
*[api/rest/property_value.md, line 73 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvV6ra-F6mBEfaGF7t:-MNZkd2J1hWx_InlPTIz:b-ug0c1m) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/api/rest/property_value.md#L73)):*
<details><summary><i>Previously, KilimAnnejaro (Anne E. Ulrich) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Could you apply it to Structure 1 as well?
___
*[api/rest/property_value.md, line 140 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvW1pV7Qm7-8gryuR0:-MNZkgji9yCpENJygV3W:b-ucfknh) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/api/rest/property_value.md#L140)):*
<details><summary><i>Previously, KilimAnnejaro (Anne E. Ulrich) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Thanks!
___
*[api/rest/property_value.md, line 479 at r9](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MNVIWV55B7r8d2rFG2X:-MNZlJQa4HvR7WRRolJw:b-92hy52) ([raw file](https://github.com/datacommonsorg/docsite/blob/c26a8475c44a4bd7a8810b13ee23c2929cd6eb15/api/rest/property_value.md#L479)):*
<details><summary><i>Previously, KilimAnnejaro (Anne E. Ulrich) wrote…</i></summary><blockquote>

I've updated the query in the most recent commit.
I am a little concerned about repeating the figure here as well as at the top, since the page seems too short for such a repetition. However, I'm not sure--thoughts?
</blockquote></details>

I think it's ok - the page is long with the example blocks. But if we wanted to include it only once, I think it belongs above, with the explanation of "direction"
___
*[api/rest/property_value.md, line 483 at r10](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MNZlYnF6kXfLt_3sPXX:-MNZlYnF6kXfLt_3sPXY:bdx4maq) ([raw file](https://github.com/datacommonsorg/docsite/blob/f2d111954a4a01040331acbce1eefb65d88eb9fd/api/rest/property_value.md#L483)):*
> ```Markdown
> 
> 
> *Figure 1. Relationship diagram for the property `containedInPlace` of the country of Argentina, for which this property takes on the value `South America`, and the city of Buenos Aires, for which this property takes on the value `Argentina`.*
> ```

What we want to highlight here is the directionality of containedInPlace. How about this:

Relationship diagram for the property `containedInPlace` of the country Argentina. Note the directionality of the property `containedInPlace`: the API returns both `in` (Buenos Aires is containedInPlace of Argentina), as well as `out` (Argentina is containedInPlace of South America).

KilimAnnejaro (commented):
> *[Reviewable](https://reviewable.io/reviews/datacommonsorg/docsite/95)* status: 3 of 7 files reviewed, 4 unresolved discussions (waiting on @beets)

___
*[api/rest/property_value.md, line 47 at r1](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvQxweTTtgqffrHi9u-r1-47:-MNa7bEPD96PHe6YWss_:b-896fix) ([raw file](https://github.com/datacommonsorg/docsite/blob/8c0b27533823a10f78fa59230102091e6fed4c38/api/rest/property_value.md#L47)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

With the suggestion in my last comment, I also wanted to clarify that this doesn't only apply to "containedInPlace" -- this applies to all properties (the second sentence of the description). Could you update that as well?
</blockquote></details>

Done.
___
*[api/rest/property_value.md, line 73 at r5](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MMvV6ra-F6mBEfaGF7t:-MNa7mgdEIZxHByHnFO-:bi5uok8) ([raw file](https://github.com/datacommonsorg/docsite/blob/401b786dc41e8475b005bdc0ef6573fe309191ac/api/rest/property_value.md#L73)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

Could you apply it to Structure 1 as well?
</blockquote></details>

Done :smiley: 
___
*[api/rest/property_value.md, line 479 at r9](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MNVIWV55B7r8d2rFG2X:-MNa8AKC6_2clKc53tkD:bo2pkvy) ([raw file](https://github.com/datacommonsorg/docsite/blob/c26a8475c44a4bd7a8810b13ee23c2929cd6eb15/api/rest/property_value.md#L479)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

I think it's ok - the page is long with the example blocks. But if we wanted to include it only once, I think it belongs above, with the explanation of "direction"
</blockquote></details>

I added it a second time. If we think that's too much, no problem--I can always just remove one of the image instances!
___
*[api/rest/property_value.md, line 483 at r10](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MNZlYnF6kXfLt_3sPXX:-MNa8O1o4UobCHoDibTZ:bk2tnxk) ([raw file](https://github.com/datacommonsorg/docsite/blob/f2d111954a4a01040331acbce1eefb65d88eb9fd/api/rest/property_value.md#L483)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

What we want to highlight here is the directionality of containedInPlace. How about this:

Relationship diagram for the property `containedInPlace` of the country Argentina. Note the directionality of the property `containedInPlace`: the API returns both `in` (Buenos Aires is containedInPlace of Argentina), as well as `out` (Argentina is containedInPlace of South America).
</blockquote></details>

YES this is exactly right! I made a few small changes to the wording and text styling, but this is much closer than where we were before.

beets (approved):
> *[Reviewable](https://reviewable.io/reviews/datacommonsorg/docsite/95)* status: 3 of 7 files reviewed, 2 unresolved discussions (waiting on @beets)

___
*[api/rest/property_value.md, line 483 at r10](https://reviewable.io/reviews/datacommonsorg/docsite/95#-MNZlYnF6kXfLt_3sPXX:-MNeYqR94ZvVuyPv0Agm:b-wdj8zi) ([raw file](https://github.com/datacommonsorg/docsite/blob/f2d111954a4a01040331acbce1eefb65d88eb9fd/api/rest/property_value.md#L483)):*
<details><summary><i>Previously, KilimAnnejaro (Anne E. Ulrich) wrote…</i></summary><blockquote>

YES this is exactly right! I made a few small changes to the wording and text styling, but this is much closer than where we were before.
</blockquote></details>

Nice, thanks!

beets:
Thanks for working through this! Let's get this in :)

PR #105, opened 2020-12-01
Python ValueError quick fix

KilimAnnejaro:
As we discussed, making a quick fix to replace the ValueError bit with NaN return. (I didn't change or update anything else in this content--only the NaN thing.)


chelseabirthday2020:
Hi @KilimAnnejaro -- this one I think can remain the same. It's the stat value endpoints that need to be changed:
https://docs.datacommons.org/api/python/stat_value.html
https://docs.datacommons.org/api/python/stat_series.html
https://docs.datacommons.org/api/python/stat_all.html

tjann:
Hi @KilimAnnejaro  -- this one I think can remain the same. It's the stat value endpoints that need to be changed:
https://docs.datacommons.org/api/python/stat_value.html
https://docs.datacommons.org/api/python/stat_series.html
https://docs.datacommons.org/api/python/stat_all.html

PR #106, opened 2020-12-01
Python quick fix

KilimAnnejaro:
Quick fixes per our conversation.


api/python/stat_all.md
```diff
@@ -36,9 +36,7 @@ with the following fields
 - `unit`: the [`unit`](https://datacommons.org/browser/unit) of the observations, if it exists.
 - `scalingFactor`: the [`scalingFactor`](https://datacommons.org/browser/scalingFactor) of the observations, if it exists.
 
-**Raises**
-
-- `ValueError` - If no statistical value found for any `Place` and `StatisticalVariable` combinations.
+If no statistical value can be found for any `Place` and `StatisticalVariable` combinations, `NaN` is returned.
```
tjann, line 39: Could you add an example here? For stat_all, I believe it returns an empty dict instead.

api/python/stat_series.md
```diff
@@ -35,9 +35,7 @@ See the [full list of StatisticalVariables](/statistical_variables.html).
 
  A `dict` mapping `date`(str) to the statistical value (float).
 
-**Raises**
-
-* `ValueError` - If no statistical value found for the place with the given parameters.
+If no statistical value can be found for the place with the given parameters, `NaN` is returned.
```
tjann, line 38: Empty dict in this case, not NaN

api/python/stat_all.md
tjann, line 39: If you run dc.get_stat_all(['badGeoId'], ['BadStaVar'])

the time series is an empty dict. I'm not really sure how to English this, but basically when there is no data for a particular Place and StatVar time series, the time series dict is empty.

tjann (commented):
Could you also update the examples on the bottom so that they return NaN / Empty Dict instead of erroring, either in this PR or a followup?

tjann (approved):
Awesome, thanks so much!

PR #109, opened 2020-12-07
Create github-pages.yml

beets:
Builds site and commits static files to gh-pages branch on every commit to master branch.

Github pages only allows a small set of plugins. We recently added jekyll-tabs in #95, and will probably add more in the future. If this action works well, will update the site settings to serve what is in gh-pages branch.


beets:
See here for more details:
https://jekyllrb.com/docs/continuous-integration/github-actions/

shifucun (approved):

PR #125, opened 2020-12-18
Sheets reorg

KilimAnnejaro:
This is my proposed plan for the Google Sheets docs:

- Migrate the sorting and place DCID files into the intro page.
- Renovate the other pages (which are all sort of logically similar) in a fashion similar to what we've been doing with the REST, Python, and Pandas docs.

This PR accomplishes the first part of the plan; it also reorganizes the video assets a little bit.


api/sheets/index.md
```diff
@@ -12,9 +12,42 @@ users to import data from the Data Commons knowledge graph. The add-on provides
...
+
+![](/assets/images/sheets/sheets_menu_bar.png)
+
+A sidebar will appear on the right of the sheet, with a search bar where you can start typing the name of the place you want. From the drop down menu, select the place you want, and its [DCID](/glossary.html) will appear in the cell that you selected.
+
+![](/assets/images/sheets/sheets_search_box.png)
+
+Note that this feature only supports places in the United States.
+
+**Sorting your results**
```
beets, line 36: Perhaps this should also be a header
```suggestion
## Sorting your results
```

beets (changes requested):
Thanks, I think this looks good

Other request is to rename the file to something with a sheets_ prefix, so we can tell these apart as we add more assets

beets (approved):

PR #130, opened 2020-12-20
Sheets get property first pass

KilimAnnejaro:
Polished first pass on Sheets get property method docs. Note: I removed a couple of the examples from the Python version of this endpoint since I am not sure if they are supported for Sheets. If this is not preferred, we can discuss and add in more examples as necessary.


api/sheets/get_property.md
```diff
@@ -6,41 +6,55 @@ parent: Google Sheets
 grand_parent: API
 ---
 
-# Get the values of a given property for a node.
+# Retrieve property values for nodes
```
beets, line 9: Should we use Retrieve for the page titles as well?

api/sheets/get_property.md
```diff
@@ -6,41 +6,55 @@ parent: Google Sheets
...
-## `=DCPROPERTY(dcids, property)`
+Given a list of nodes and a property label, returns values associated with the given property for each node.
 
-Given a property, together with a single DCID, a row of DCIDs, or a column of DCIDs, get the values of the given property for those DCIDs.
+## General information about this endpoint
 
-**Arguments**
-*    `dcids` - DCIDs to get the property for
-*    `property` - property to get
+**Signature**: `=DCPROPERTY(dcids, property)`
```
beets, line 15:
```suggestion
**Formula**: `=DCPROPERTY(dcids, property)`
```

api/sheets/get_property.md
```diff
@@ -6,41 +6,55 @@ parent: Google Sheets
...
 
-**Returns**
+**Required arguments**:
 
-The properties of the DCIDs. For a single DCID, the result is a column of the given property's values for that DCID. For a row of DCIDs, the result is a matrix where each column contains the values of the given property for the DCID at the column's index. For a column of DCIDs, the result is a matrix where each row contains the values of the given property for the DCID at the row's index.
+*   `dcids`: A list of nodes to query, identified by their Data Commons identifiers.
+*   `property`: The property to query for.
```
beets, line 20: Should we keep the "Returns" section here?

api/sheets/get_property.md
```diff
@@ -6,41 +6,55 @@ parent: Google Sheets
...
-## Examples
+## Assembling the information you will need for a call to this method
 
-In this example, we find the `containedInPlace` of the Data Commons nodes for [California](https://datacommons.org/browser/geoId/06) and [Mountain View](https://datacommons.org/browser/geoId/0649670). The function call in the image uses a single call to get the names of both nodes, but it is equivalent to typing:
+Going into more detail on how to assemble the values for the required arguments:
 
-```
-=TRANSPOSE(DCPROPERTY("geoId/06", "containedInPlace"))
-```
+ - `dcids`: Data Commons uniquely identifies nodes by assigning them DCIDs, or Data Commons IDs. Your query will need to specify the DCIDs for the nodes of interest. More information about DCIDs is available in [the glossary](/glossary.html).
```
beets, line 26: Nice - can we keep the link to the glossary to the other sheet API pages?

api/sheets/get_property.md
```diff
@@ -6,41 +6,55 @@ parent: Google Sheets
...
 
-The properties of the DCIDs. For a single DCID, the result is a column of the given property's values for that DCID. For a row of DCIDs, the result is a matrix where each column contains the values of the given property for the DCID at the column's index. For a column of DCIDs, the result is a matrix where each row contains the values of the given property for the DCID at the row's index.
+*   `dcids`: A list of nodes to query, identified by their Data Commons identifiers.
+*   `property`: The property to query for.
 
-## Examples
+## Assembling the information you will need for a call to this method
 
-In this example, we find the `containedInPlace` of the Data Commons nodes for [California](https://datacommons.org/browser/geoId/06) and [Mountain View](https://datacommons.org/browser/geoId/0649670). The function call in the image uses a single call to get the names of both nodes, but it is equivalent to typing:
+Going into more detail on how to assemble the values for the required arguments:
```
beets, line 24: Should we add a note here about visiting the browser page for a node to see available properties?

beets (changes requested):
thanks!

I'm looking at the old note about TRANSPOSE, and I wonder if it's worth adding a more complex example (that uses columns, or maybe a table that has multiple properties as columns?) wdyt?

api/sheets/get_property.md
KilimAnnejaro: I think I mentioned this in another PR, but I'm not sure if the change would negatively affect the user experience on the sidebar due to the increased width.

api/sheets/get_property.md
KilimAnnejaro: I don't understand?

api/sheets/get_property.md
KilimAnnejaro: Yes, this is an excellent idea!

api/sheets/get_property.md
KilimAnnejaro: I added a comment towards the end.

api/sheets/get_property.md
beets: I think I meant that this sentence is a nice clear link to the glossary:
"More information about DCIDs is available in [the glossary](/glossary.html)."

and wondering if we should add that to the other sheets API pages as well

api/sheets/get_property.md
beets: +1

api/sheets/get_property.md
```diff
@@ -6,41 +6,63 @@ parent: Google Sheets
...
-```
+ - `dcids`: Data Commons uniquely identifies nodes by assigning them DCIDs, or Data Commons IDs. Your query will need to specify the DCIDs for the nodes of interest. More information about DCIDs is available in [the glossary](/glossary.html).
 
-and
+ - `property`: The property whose value you are interested in, such as "name" for the name of a node, or "typeOf" for the type of a node. If you aren't sure what properties are available for a particular DCID, you can use the [Data Commons graph browser](https://datacommons.org/browser/) to look up the DCID of interest and see what properties it is associated with.
 
-```
-=TRANSPOSE(DCPROPERTY("geoId/0649670", "containedInPlace"))
-```
+ **Returns**
```
beets, line 30: Does the returns section belong above the "assembling" section, along with the formula signature and arguments?

api/sheets/get_property.md
```diff
@@ -6,41 +6,63 @@ parent: Google Sheets
...
-### Input
+>  **NOTE:**
+>
+>  It's best to minimize the number of function calls to `DCPROPERTY` by using a single call to get a variable for a row/column of places and/or a column/row of times. This is because a spreadsheet will make one call to a Google server [per custom function call](https://developers.google.com/apps-script/guides/sheets/functions#optimization). If your sheet contains thousands of separate calls to `DCPROPERTY`, expect it to be slow.
 
-![](/assets/images/sheets/sheets_get_property_input.png)
+## Example requests and responses
 
-### Output
+### Example 1: Retrieve the common names of the country of Côte d'Ivoire.
```
beets, line 40: same comment applies from a previous PR (I think we renamed this to "retrieve the common names of a country by it's dcid"

api/sheets/get_property.md
KilimAnnejaro: Yep, we can (and will) incorporate this in other places!

api/sheets/get_property.md
KilimAnnejaro: We haven't usually done something like that (by analogy see https://docs.datacommons.org/api/rest/property_value.html), but if we want to move in that direction we can.

api/sheets/get_property.md
KilimAnnejaro: updated!

api/sheets/get_property.md
beets: could we start by moving this one up? this is more of a summary compared to the rest example.

api/sheets/get_property.md
beets: great thanks!

beets (commented):
thanks for the updates, just a couple more small comments

api/sheets/get_property.md
KilimAnnejaro: sure, no problem

KilimAnnejaro:
Wrt comment about transpose: for examples with multiple columns, isn't that being worked on in tutorials? Also, not sure what is meant by "example that uses columns."

beets:
I meant that perhaps we should have an example where the data is laid out in columns (so the formula would select a row instead). I think what we have is good though, and as you mentioned, we have those examples elsewhere.

beets (approved):

PR #102, opened 2020-11-29
Stat value REST docs update

KilimAnnejaro:
Updating the REST docs for the stat value endpoint.

---
This change is [Reviewable](https://reviewable.io/reviews/datacommonsorg/docsite/102)

PR #134, opened 2021-01-07
Glossary entry for the observation period

KilimAnnejaro:
Per discussion on https://github.com/datacommonsorg/docsite/pull/102, updating the glossary a bit.


beets (approved):
Thanks!

PR #138, opened 2021-01-09
Small fix on REST docs

KilimAnnejaro:
Needed to ensure uniform capitalization across docs with tabs plugin to avoid too many JS tabs appearing.


beets (approved):
thanks for the cleanup!

Issue #148, opened 2021-02-17
API doc for new REST endpoint: /place/stat-vars


beets:
This endpoint allows a user to explore statistical variables which are available for a set of places. Available as both GET and POST

Endpoint: /place/stat-vars ([code](https://github.com/datacommonsorg/mixer/blob/master/proto/mixer.proto#L760))
Parameters: dcids (list)
Request: https://api.datacommons.org/place/stat-vars?dcids=country/USA
Response: {"places":{"country/USA":{"statVars":["dc/zl73qp466bs28","dc/z6jy58c28k7zh"]}}}

Example: http://api.datacommons.org/place/stats-var?dcids=geoId/06&dcids=zip/94025

beets:
the link to the code above is how we are specifying the api on the server. i added it for reference, though we don't need to include it in the final published doc.

PR #156, opened 2021-02-27
REST place stat vars endpoint

KilimAnnejaro:
Resolves https://github.com/datacommonsorg/docsite/issues/148.


beets (commented):
thanks! looks great, only nit is that tabs.js doesn't need to be added explicitly anymore since that was added to the base template.

beets (approved):

beets:
thx!!

PR #186, opened 2022-01-07
Update env for deploy action

beets:
Actions weren't running because the requested environment was obsolete. Also pull latest version of the action


shifucun (approved):
thanks for fix!

beets:
I'm not sure how to test before merging -- hope it works! if not there will be followup pr's :)

PR #192, opened 2022-01-19
fix alphabetical order for data sources

n-h-diaz:
will follow up with table of contents
let me know when it looks okay to add redirects :)


pradh:
> will follow up with table of contents let me know when it looks okay to add redirects :)

Maybe after TOC. @beets wdyt?

beets:
> > will follow up with table of contents let me know when it looks okay to add redirects :)
> 
> Maybe after TOC. @beets wdyt?

Yup, that sgtm!

PR #194, opened 2022-01-20
add toc and expose data sources pages

n-h-diaz:
will add redirects on website as follow up

### DIFF
--- a/datasets/biomedical.md
+++ b/datasets/biomedical.md
@@ -8,7 +8,7 @@ parent: Data Sources
 {:.no_toc}
 # Biomedical Data
 
-1. TOC
+* TOC
 {:toc}
 
 #### [ChEMBL](https://www.ebi.ac.uk/chembl/)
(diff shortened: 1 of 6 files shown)


datasets/biomedical.md
```diff
@@ -5,8 +5,12 @@ nav_order: 3
 parent: Data Sources
 ---
 
+{:.no_toc}
 # Biomedical Data
 
+1. TOC
```
beets, line 11: could we use disc bullets for these? so:
```suggestion
* TOC
```

beets (approved):
thanks - looks good! 

it could use a big of styling though.. how about:

```css
#markdown-toc {
    background: var(--light);
    padding: 1rem 2rem;
    border-radius: 5px;
}
```

datasets/biomedical.md
n-h-diaz: done

n-h-diaz:
thanks for the review!

PR #215, opened 2022-05-06
update data sources pages

n-h-diaz:


scripts/generate_datasets.sh
```diff
@@ -0,0 +1,15 @@
+# Script to copy generated dataset pages to the local docsite repo.
```
pradh, line 1: I wonder if this script should be in g3?  Given its reference to blaze run etc.

For example, the script gets run from a g4 client, takes the github docsite dir as argument, does the blaze run to produce stuff in /tmp/dataset_info, copies over the *.md to github dir, prepares a commit with the modified files, and informs user to review commit and send out PR (like https://github.com/datacommonsorg/mixer/blob/master/scripts/prepare_release.sh#L60-L62).

datasets/index.md
```diff
@@ -7,12 +7,17 @@ has_children: true
 
 # Data Sources
 
-Data in the Data Commons Graph comes from a variety of sources, each of which often includes multiple surveys. Some sources/surveys include a very large number of variables, some of which might not yet have been imported into Data Commons. The sources have been grouped by category and are listed alphabetically within each category.
+Data in the Data Commons Graph comes from a variety of sources, each of which often includes multiple surveys. Some sources/surveys include a very large number of variables, some of which might not yet have been imported into Data Commons. The sources have been grouped by vertical and are listed alphabetically within each vertical.
```
pradh, line 10: nit: Wonder if "category" is the more user-friendly term?

scripts/generate_datasets.sh
```diff
@@ -0,0 +1,15 @@
...
+then
+  echo "Please rerun 'blaze run //datacommons/import/mcf/dataset_info:parse_dataset_info' from a CitC client to generate the dataset pages."
+else
+  # Keep Upcoming Data Inmports page, which is manually updated.
+  find "./datasets/" -type f -not -name "*upcoming*" -delete
+
+  for file in "/tmp/dataset_info/"*
+    do
+      cp $file "./datasets/"
+    done
```
pradh, line 13: nit: why not `cp /tmp/dataset_info/*.md $DST/datasets/`

pradh (approved):
Very nice!

It might be useful to add a comment that this change is expected to include *all* existing documentation, but has been laid out as verticals and generated automatically from the manifest.

scripts/generate_datasets.sh
n-h-diaz: sounds good, removed and will add a script to g3

datasets/index.md
n-h-diaz: done

beets (commented):
thanks for the update!

for the moved / deleted files, we should add redirects (either to new url, like biomedical, or back to the index page for straggler links out there)

beets:
> Very nice!
> 
> It might be useful to add a comment that this change is expected to include _all_ existing documentation, but has been laid out as verticals and generated automatically from the manifest.

could you update the PR description with this? thanks!

PR #227, opened 2022-08-04
Fix broken courseware page link.

juliawu:
Fixes the link to the courseware page in our contributing page that currently leads to a 404 error.


chejennifer (approved):
thanks for the fix!

shifucun:
Just a minor comment. For a PR, it's good to start from a clean new local branch so the commit history only contains the related change for this PR.

juliawu:
> Just a minor comment. For a PR, it's good to start from a clean new local branch so the commit history only contains the related change for this PR.

Thanks for the tip, will do for future PRs!

PR #245, opened 2022-09-09
Some restructure of V1 API home page

shifucun:
* Have simple and bulk versions together under the same section
* Shuffle the groups
* Shortened descriptions


shifucun:
I use [bulk] but I guess there should be better css treatment for it, maybe a label?

juliawu (approved):
Great idea! I'll add buttons in CSS later.

PR #247, opened 2022-09-09
[v1 API Docs] Add css for bulk tags, plus some quick fixes

juliawu:
Convert [bulk] on api/rest/v1/index.md to bulk tags that render with CSS.

Also corrects some minor formatting errors in the getting started guide.

Screen Shot 2022-09-09 at 12 51 03 PM


api/rest/v1/index.md
```diff
@@ -33,7 +33,7 @@ https://api.datacommons.org
 ## Simple vs Bulk Query
```
shifucun, line 33: Should all these titles use "###" instead of "##"?

shifucun (approved):
looks very nice! Now I know how to add a css tag :)

shifucun:
just wonder if the v0 folder change affects current deployment? anywhere else need to be updated to refer to the files?

juliawu:
It doesn't change the navigation behavior, but does change the link. Reverted the v0 folder change for now.

shifucun (approved):

PR #262, opened 2022-09-14
[v1 API Docs] Fix (!) Icon & CSS refactoring

juliawu:
* Refactors some css by adding scss variables
* Resizes the (!) Icon in important alert boxes to be the same height as the text
Current:
image
Proposed:
image


_sass/api_documentation.scss
```diff
@@ -230,7 +230,7 @@ bulk-tag {
 .exclamation-icon {
   background: var(--dc-red-lite);
   color: white;
-  font-size: 1em;
+  font-size: 0.8em;
```
beets, line 233: Ah, there was also this issue with the "see also" icon. Worth extracting out some common css rules between the two.

beets (approved):

PR #266, opened 2022-09-15
[blog] set max image width

chejennifer:


_sass/blog.scss
```diff
@@ -15,4 +15,8 @@
   color: var(--dc-gray);
   text-transform: uppercase;
   font-size: 1rem;
+}
+
+img {
```
beets, line 20: Can we restrict this to blog pages? There are other images with text (e.g. tutorials) which would be too small to read if we restrict the width.

beets (commented):
Also, sorry for not catching this earlier, but could you also update the blog post image to be from a smaller window width and height.

_sass/blog.scss
chejennifer: done

chejennifer:
> Also, sorry for not catching this earlier, but could you also update the blog post image to be from a smaller window width and height.

what do you mean by updating the image to be from a smaller window width and height? like take a new screenshot? what size are you thinking

beets:
> > Also, sorry for not catching this earlier, but could you also update the blog post image to be from a smaller window width and height.
> 
> Do you mean take a new screenshot for the blogpost? what size are you thinking

Yup - new screenshot. Maybe 960 (w/o stat var panel) or 1080 w/ stat var panel

chejennifer:
> > > Also, sorry for not catching this earlier, but could you also update the blog post image to be from a smaller window width and height.
> > 
> > 
> > Do you mean take a new screenshot for the blogpost? what size are you thinking
> 
> Yup - new screenshot. Maybe 960 (w/o stat var panel) or 1080 w/ stat var panel

updated

beets (approved):
thx!

PR #267, opened 2022-09-16
fix dropdown naming inconsistency

chejennifer:
- the dropdown on website has the data download tool option as "Data Download Tool"


juliawu (approved):
LGTM!

PR #275, opened 2022-11-01
Fix broken link

beets:
Current prod link goes to https://docs.datacommons.org/api/rest/v1/property/values/linked which is broken.


juliawu (approved):
Good catch! Thanks for the fix!

PR #276, opened 2022-11-01
fix broken example query

beets:
there's an extra get param in the triples example


juliawu (approved):
Thanks for the catch and fix!

Issue #280, opened 2022-11-18
Bulk tags show up in page titles


beets:
We should perhaps add a property to the page (similar to page-order, etc), that can be used in the sidebar generator to add the tag there. Optionally could be used to add the word Bulk to the page title as well.

![image](https://user-images.githubusercontent.com/6052978/202796354-812a0bb3-ae66-4ee2-ae6e-11951cf9b980.png)

juliawu:
Fixed by #282 and #283!

PR #282, opened 2022-11-30
Fix bulk-tags showing up in page titles

juliawu:
Fixes issue #280 

This PR:
- adds a `is_bulk` property to v1 bulk api pages
- Adds logic to the sidebar to render bulk-tags
- Adds "(Bulk version)" to breadcrumbs of bulk pages
- Removes bulk tags from page titles

Screen Shot 2022-11-30 at 3 26 29 PM

Notice tab heading no longer leads with html tags

### DIFF
--- a/api/rest/v1/bulk_observations_series.md
+++ b/api/rest/v1/bulk_observations_series.md
@@ -1,12 +1,12 @@
 ---
 layout: default
-title: <bulk-tag>BULK</bulk-tag> Series of Observations
+title:  Series of Observations
 nav_order: 108
 parent: REST (v1)
 grand_parent: API
 published: true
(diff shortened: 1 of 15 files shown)


api/rest/v1/bulk_observations_series.md
```diff
@@ -1,12 +1,12 @@
 ---
 layout: default
-title: <bulk-tag>BULK</bulk-tag> Series of Observations
+title:  Series of Observations
```
shifucun, line 3: wonder if the extra space before "Series" matters?

juliawu: Don't think it matters, but cleaned up the extra space!

PR #284, opened 2022-12-03
Updating Data Sources pages

chejennifer:


n-h-diaz (approved):

datasets/Environment.md
```diff
@@ -45,6 +45,13 @@ Data Commons includes relative measures of risk from the 18 natural hazards incl
 A global inventory of glaciers, including surface areas.
 
 
+### [Google](https://www.google.com/)
```
juliawu, line 48: What are we sourcing from Google?

n-h-diaz: it looks like this is the Inferred Floods from Dynamic World dataset, coming from https://developers.google.com/earth-engine/datasets/catalog/GOOGLE_DYNAMICWORLD_V1

beets: Can we title this as Dynamic World Project instead? Let's also make sure that we include the required attribution from https://developers.google.com/earth-engine/datasets/catalog/GOOGLE_DYNAMICWORLD_V1#terms-of-use

chejennifer: It looks like the title is the dataset source name & there are other datasets under the Google dataset source, so I'm not sure changing the title makes sense: https://source.corp.google.com/piper///depot/google3/datacommons/import/mcf/manifest/prod_dataset_info.textproto;l=184;bpv=1

Created a CL to add the required attribution: https://critique.corp.google.com/cl/493312619

chejennifer: Updated

chejennifer: Updated

beets (approved):
thanks for the updates!

PR #288, opened 2022-12-29
Add a 3rd folder level to sidebar

juliawu:
This PR allows the sidebar to support one additional layer of folder nesting, which we need to add the new Data Literacy courseware to the website.

Now a structure like
```
Folder 1
  |__ Folder 2
    |__ Folder 3
      |_ my nested page 
    
```

Can be created using the following yaml front matter:
```
---
parent: Folder 3
grand_parent: Folder 2
great_grand_parent: Folder 1
---
```


jehangiramjad (approved):
Approving to unblock. But @chejennifer might have comments which we should address.

chejennifer (approved):
this is cool!

juliawu:
Added one more change to fix breadcrumbs for 4th level nested files.

Issue #292, opened 2023-02-02
Call out observation series API data is sorted by time


shifucun:
Example page now: https://docs.datacommons.org/api/rest/v1/bulk/observations/series

We should call mention this in all series API

PR #293, opened 2023-02-03
[v1 api docs] clarify series are sorted by time.

juliawu:
Adds a clarification to bulk series pages that observations are returned in chronological order to the response fields.

This fixes issue #292 

Confirmed simple series page already mentions chronological order, so no change is needed on /v1/observations/series.

Example:
Screenshot 2023-02-03 at 8 58 01 AM


shifucun (approved):
thanks!

PR #294, opened 2023-02-03
Remove dangling biomeidcal md files

shifucun:
"biomedical.md" and "Biomedical.md" causes trouble for git to work on Mac.

Are the old links being used somewhere or safe to remove?

### DIFF
--- a/Biomedical.md
+++ b/Biomedical.md
@@ -1,4 +1,0 @@
----
-layout: redirect
-redirect: /datasets/Biomedical.html
----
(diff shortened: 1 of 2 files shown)


pradh (approved):
Maybe double check with Natalie.

juliawu:
While here or in a followup PR, would also be good to remove one of `datasets/Biomedical.md` and `datasets/biomedical.md` which I think are copies of each other. But @n-h-diaz might have more context.

n-h-diaz (approved):

PR #310, opened 2023-03-04
Update nav order for documents

lucy-kind:
Data Imports, then Datasets, then Glossary


beets (approved):
thanks for the quick fix!

PR #315, opened 2023-03-22
Correct syntax of v1/bulk/info/place

### LINKED ISSUE #314: Incorrect request/response fields in docs for /v1/bulk/info/place ?
https://docs.datacommons.org/api/rest/v1/bulk/info/place

There are references to `entities` / `entity`, but looks like the implementation uses `nodes` / `node`.

juliawu:
Replace 'entities' with 'nodes' in the v1/bulk/info/place API

The current documentation page for v1/bulk/info/place uses the incorrect query parameter 'entities'.
This PR updates the documentation page to use the correct query parameter 'nodes'.

This PR fixes #314


pradh (commented):
Thanks for the fix!

I didn't check if the non-bulk version (/v1/info/place) is fine, is it?

juliawu:
@pradh Non-bulk version is fine! Non-bulk doesn't use named query parameters, nodes are passed in as part of the path.

PR #316, opened 2023-03-22
Update custom DC doc based on user feedback

shifucun:


custom_dc/upload_data.md
```diff
@@ -76,11 +76,14 @@ information import is untouched.
 
 ## Storage Layout
 
-All custoom Data Commons data are stored under one GCS folder. Below shows a
+All custom Data Commons data are stored under one GCS folder. Below shows a
 typical layout.
 
+Note, create a "root_folder" under the desired GCS bucket, which will be used to
```
juliawu, line 82: Does the root folder have to be named `root_folder` or can the name be anything?

If the root folder needs a specific name, let's specify the name here.
If the name can be anything, replace "root_folder" with the words root folder (no quotes) to reduce confusion.

shifucun: good suggestion. updated.

juliawu (approved):
Thanks for the updates!

PR #330, opened 2023-07-18
Update colors

chejennifer:
- update link and header colors to match dc website


juliawu (approved):
Thanks for the quick fix!

PR #340, opened 2023-09-05
DC API v2 doc updates

donaldrgosselin:


api/rest/v2/node.md
```diff
@@ -158,6 +159,7 @@ Response:
   }
 }
 ```
+
```
shifucun, line 162: please remove the empty line here. This will break the appearance.  Same for below

api/rest/v2/node.md
```diff
@@ -56,26 +56,27 @@ X-API-Key: {your_api_key}
 
 JSON Data:
 {
-  "nodes": [
-      "{value_1}",
-      "{value_2}",
-      ...
-    ],
-  "property": "{property_expression}"
+"nodes": [
```
shifucun, line 59: Can you revert this? I think this would be from the auto format from vscode. You may want to turn that off.

api/rest/v2/index.md
```diff
@@ -30,3 +51,84 @@ The base URL for all endpoints below is:
...
+
+```
+country/USA<-containedInPlace+{typeOf: City}
+```
+
+This makes it clear that nodes with “typeOf: City” are queried and are linked to “country/USA” by one or more “containedInPlace” properties.
+
+As an example of property chaining, consider the following diagram:
+
+![alt text](image.png)
```
shifucun, line 100: should check in or point to the correct path of image.png.

Can you double check that all the link and text in this page works.

donaldrgosselin: I checked in the image and double checked the links.

beets:
hi don - can you upload your local changes that addresses the other comments in the PR? thanks

donaldrgosselin:
Sorry, I'm not sure what you're referring to. I stopped working on the
markdown files yesterday and am now focusing on the Google Docs file.

beets:
closing this PR for now, since we have been updating the content in subsequent PRs from the doc

PR #341, opened 2023-09-05
Updated docsite header, footer, fonts, and link colors to match new site design

dwnoble:
### Header before:

Screenshot 2023-09-05 at 2 57 06 PM

### Header after
Screenshot 2023-09-05 at 2 57 10 PM

### Header before (responsive)
Screenshot 2023-09-05 at 2 57 51 PM

### Header after (responsive)
Screenshot 2023-09-05 at 2 58 00 PM

### Footer before
Screenshot 2023-09-05 at 2 57 31 PM

### Footer after
Screenshot 2023-09-05 at 2 57 21 PM

### New footer just before responsive breakpoint
Screenshot 2023-09-05 at 2 59 43 PM

### New footer just after responsive breakpoint

Screenshot 2023-09-05 at 2 59 50 PM


shifucun:
We prob want to remove the search box in the header?

dwnoble:
> We prob want to remove the search box in the header?

We could, but the search might still be useful for searching doc specific terms:
Screenshot 2023-09-05 at 4 33 23 PM

shifucun:
> > We prob want to remove the search box in the header?
> 
> We could, but the search might still be useful for searching doc specific terms: Screenshot 2023-09-05 at 4 33 23 PM

It's just there is abrupt header change when click on the link from main page to the doc page, which is a bit surprising to users. To make them as seamlessly one site, I still feel removing the search is nice.

dwnoble:
That's fine with me- @beets wdyt?

beets (approved):
this is really awesome -- thanks dan!

i would leave the searchbox for now. we refer to it in a few places in our docs, and there are many other changes to the site when you switch over to here from the main site, so i think it's not too bad leaving that there.

one request before merging is to keep the red DC logos since we autopush on green. we can switch it out to blue next week.

dwnoble:
Sounds good- i switched the logo/favicon to red and am leaving the search bar in place for now

PR #362, opened 2023-09-15
Always open all the sidebar menu

shifucun:
This is to revert an early change to collapse some of the menu on page loading.

This does not work well when opening a top level item, and click on its child. Then the page reload with the top level item closed again.

We strongly need a simple state management of the site. Given all the complexity of Jekyll, we should consider using React to re-write this site.


pradh (approved):

shifucun:
thanks!

PR #363, opened 2023-09-26
Fix broken links in courseware blog post

juliawu:
Updates the links in the older courseware blog post to point to the correct URLs, since file paths have been changed since the post was written.


pradh (approved):
Thanks for the fix!

PR #368, opened 2023-09-28
Updating Data Sources pages

chejennifer:


datasets/index.md
```diff
@@ -1,9 +1,8 @@
 ---
 layout: default
 title: Data Sources
-nav_order: 110
+nav_order: 9
```
beets, line 4: I believe @shifucun made these very far apart so it's easier to slot in new pages, just like freeway exit numbers.

beets (approved):
thanks for the update!

datasets/index.md
shifucun: Yep! I wonder how are these auto-generated? We may want to fix there.

chejennifer: ok, I reverted for now, but yes should probably fix in the auto-generation

PR #374, opened 2023-11-02
Update ruby, bundle and dependencies

beets:
I kept running into issues after os updates with ruby 3.x (it hangs on `bundle update` and fails to run `bundle exec jekyll serve`. I have ruby 3.2.2 locally.
First deleted the lines "BUNDLED WITH" in Gemfile.lock
Then ran the following commands to update:

```
bundle update --ruby --bundler
bundle update --verbose
bundle add webrick
```

Also had to add webrick to the bundle due to this issue: https://github.com/jekyll/jekyll/issues/8523


beets:
Curious if either of you have run into this issue before. In the past I would have had to downgrade ruby as a workaround.

shifucun:
> Curious if either of you have run into this issue before. In the past I would have had to downgrade ruby as a workaround.

I have run into some bundle issue before but can not remember exactly. Good to update though!

shifucun (approved):

PR #377, opened 2023-11-07
Updating Data Sources pages

pradh:


datasets/index.md
```diff
@@ -3,7 +3,6 @@ layout: default
 title: Data Sources
 nav_order: 110
 has_children: true
-has_toc: false
```
beets, line 6: is this an intentional change?

beets (approved):

datasets/index.md
pradh: Oh oops, no, it is what was generated by the script.  Should I revert this?

beets: hm, looks like there is no difference in output. it would make sense to be false by default so this is ok to remove

PR #384, opened 2024-02-20
FEPRA Compliance for courseware

jehangiramjad:
Adding Legal approved FEPRA compliance statements to the course/curriculum pages.


juliawu (approved):
nit: For all of these, could we make "FERPA" in the first sentence a link to a page explaining what it is? Perhaps https://www2.ed.gov/policy/gen/guid/fpco/ferpa/index.html

PR #390, opened 2024-04-30
Add a place types page

juliawu:
This PR:
* Adds a page listing available place types.
* Adds links throughout our API docs to the new page.

Context: Often our APIs ask users to provide a contained place type (ex: “childPlaceType” for the [Map web component](https://docs.datacommons.org/api/web_components/map) or “typeOf” for a [REST v2 containedInPlace query](https://docs.datacommons.org/api/rest/v2/observation)). However, we don’t make the list of available place types to use with those APIs available to users beyond our [Knowledge Graph](https://datacommons.org/browser/Place), but even then the user would need to understand our graph structure and schema to figure out which table in the knowledge graph browser to look at.

Screenshot of the new page:

localhost_4000_place_types html (1)


kmoscoe:
- Capitalize Statistical Variable Explorer.
- what about stuff like geoId and latLng and those? Are they not also place types? (excuse my ignorance, I'm frankly really confused about all this schema stuff :-)
- Generally all the headings should use sentence case, not title case, if we were properly following the Google style guide. But we don't -- notwithstanding what the README.md in the docsite says. :-)

juliawu:
> * Capitalize Statistical Variable Explorer.

Done!

> * what about stuff like geoId and latLng and those? Are they not also place types? (excuse my ignorance, I'm frankly really confused about all this schema stuff :-)

Not place types.

The technical definition is that a place type needs to be (1) a node in the graph and (2) a `subClassOf` the [Place node](https://datacommons.org/browser/Place) (see the Subject: Class table. If you click into any of those nodes, entries in their Subject: Class tables are also place types.).

"geoId" and "latLng" are common prefixes we see in dcids we happened to use for our convenience, but they are not nodes themselves. "geoId"s I believe we [took from the US Census](https://www.census.gov/programs-surveys/geography/guidance/geo-identifiers.html) when creating dcids identifiers for US places.

> * Generally all the headings should use sentence case, not title case, if we were properly following the Google style guide. But we don't -- notwithstanding what the README.md in the docsite says. :-)

Just made the headings on the new page sentence case, and updated the screenshot.

kmoscoe:
Thanks so much for the explanation!

kmoscoe (approved):

PR #399, opened 2024-05-17
Update troubleshooting.md

kmoscoe:


custom_dc/troubleshooting.md
```diff
@@ -1,4 +1,4 @@
----
+s---
```
juliawu, line 1: Take out leading s

juliawu (approved):
Thanks for the fix!

custom_dc/troubleshooting.md
kmoscoe:
```suggestion
---
```

PR #404, opened 2024-05-20
added nice TOC to all pages and fixed top-level heading tag

kmoscoe:


juliawu (approved):

juliawu (approved):
Thanks for adding TOC!

PR #418, opened 2024-06-06
Fix typo in custom_dc overview

juliawu:
First line in index.md uses a final comma instead of a period.


kmoscoe (approved):
Gaahh!  Thanks for catching, looks like a weird Github syncing error. :-(

PR #420, opened 2024-06-07
Remove markdown formatting from link in code block

hqpho:


keyurva (approved):
Your first PR!

kmoscoe (commented):
Are you sure it works? I've found that bare links don't always work for me. (They should, but the Github pages are wonky.)

hqpho:
> Are you sure it works? I've found that bare links don't always work for me. (They should, but the Github pages are wonky.)

Yes, I'm sure! This link is in a code block with instructions to copy/paste it, so it needs to be formatted differently from other links on the page. Without this change, it isn't valid as a copy/pasted URL.

PR #429, opened 2024-06-27
Try to fix image link in /index.md

beets:


kmoscoe (approved):
Not sure this is going to help, as there are other places where : width works, e.g. https://docs.datacommons.org/custom_dc/quickstart.html, but let's give it a try.

beets:
yes, the next thing i'll try is moving the image into a subfolder

kmoscoe:
Seems to still not be working. :-(

PR #444, opened 2024-07-09
move template file to where it should be and update formats

kmoscoe:


kmoscoe:
This is a non-published file intended to help with API doc-writing, shows how to use the stylesheet (referenced from https://github.com/datacommonsorg/docsite/blob/master/_sass/api_documentation.scss) and avoid things like PR 441. :-)

PR #453, opened 2024-07-22
Bigquery cleanup

kmoscoe:
This PR cleans up the BigQuery docs by:
1) Removing redundant pages about basic DC info, now replaced by updated Key Concepts page
2) Redirecting extraneous pages to appropriate places
3) Rewriting the landing page
4) Applying minor edits
5) Putting a more descriptive heading


bigquery/query_places.md
```diff
@@ -64,13 +63,13 @@ WHERE
...
 ```
 
-## Entities of a certain type in a Place
+## Entities of a certain type in a place
 
-This is the same sequence as above, except replace place with entities of type X, where X could be School, Facility, etc.
+This is the same sequence as above, except replacing place with entities of type X, where X could be school, facility, etc.
 
-### List all entities of non-place types (School, Facility, etc.) contained in Y
+### List all entities of non-place type contained in Y
```
juliawu, line 70: Can leave for a followup PR:

This particular section is confusing for me. What is a non-place type? If it refers to facilities like in the example, I think it's fair for users to think a facility is a type of place (as opposed to say, an animal species). It might be helpful to expand on this slightly.

juliawu (approved):

bigquery/query_places.md
kmoscoe: OK, perhaps for another PR. https://docs.datacommons.org/place_types.html does not list schools or facilities as places. IIUC, these types of entities are top-level entities, i.e. types of class, e.g. https://screenshot.googleplex.com/wjsuyPbuHsMpG4R. Perhaps I could add something about this to https://docs.datacommons.org/data_model.html#entity, and link to that?

PR #455, opened 2024-07-22
Updating Data Sources pages

chejennifer:


datasets/index.md
```diff
@@ -20,23 +20,4 @@ Data in the Data Commons Graph comes from a variety of sources, each of which of
...
-This chart goes a level deeper and illustrates the total number of statistical variables available at the state level for each country worldwide.
-<p> &nbsp; </p>
-<div class="flourish-embed flourish-map" data-src="visualisation/17716780"></div>
-<p> &nbsp; </p>
-
-Finally, this third chart illustrates the total number of statistical variables available at the district/county level worldwide.
-
-<p> &nbsp; </p> 
-<div class="flourish-embed flourish-map" data-src="visualisation/17719383"></div>
-<p> &nbsp; </p>
```
juliawu, line 42: Is this deletion intentional? I think having these visualizations may have been an ask from James. @miss-o-soup

chejennifer: good catch, I just ran the script & I don't think this should be deleted. Let me update

chejennifer: done

juliawu (approved):
Thanks for updating!

PR #467, opened 2024-08-05
Add some more census.gov types

kmoscoe:


place_types.md
```diff
@@ -43,17 +43,19 @@ countries, but not all countries.
 |[Village](https://datacommons.org/browser/Village)|A small clustered human settlement smaller than a town. |
 {: .doc-table}
 
-### US-specific administrative divisions
+### U.S.-specific administrative divisions
```
pradh, line 46: I'm not sure if tract is considered an administrative division.  Perhaps "U.S.-specific place types"?

place_types.md
```diff
@@ -43,17 +43,19 @@ countries, but not all countries.
...
 
 These place types can only be used for places that are contained within the
 [United States](https://datacommons.org/place/country/USA) (DCID:
-[country/USA](https://datacommons.org/browser/country/USA)).
+[country/USA](https://datacommons.org/browser/country/USA)). See https://datacommons.org/browser/dc/base/BaseGeos for additional places defined for the U.S.
```
pradh, line 50: I slightly worry about the readability of the Browser page.  @miss-o-soup?

place_types.md
```diff
@@ -43,17 +43,19 @@ countries, but not all countries.
...
-|[CensusZipCodeTabulationArea](https://datacommons.org/browser/CensusZipCodeTabulationArea)|US zip codes as defined by the US Census Bureau.<br><br>For example, [94043](https://datacommons.org/place/zip/94043).<br><br>While there is significant overlap, these codes don't always correspond to the zip codes used by the US Postal Service.|
+|[State](https://datacommons.org/browser/State)|U.S. states.<br><br>For example, [California](https://datacommons.org/place/geoId/06) or [Maryland](https://datacommons.org/place/geoId/)|
+|[County](https://datacommons.org/browser/County)|U.S. counties.<br><br>For example, [Santa Clara County](https://datacommons.org/place/geoId/0669084)|
+|[CensusBlockGroup](https://datacommons.org/browser/geoId/01003990000)|U.S. block groups as defined by the U.S. Census Bureau.<br><br>For example, [Block Group 0](https://datacommons.org/browser/geoId/010039900000)|
+|[CensusTract](https://datacommons.org/browser/CensusTract)|U.S. census tracts as defined by the U.S. Census Bureau.<br><br>For example, [Census Tract 10](https://datacommons.org/browser/geoId/01015001000)|
```
pradh, line 57: If we want to order them from potentially larger place-types to smaller, then the order would be:
- State
- County
- CensusZipCodeTabulationArea
- CensusTract
- CensusBlockGroup

(counties contain tracts, tracts contain block-groups)

pradh (approved):
Thanks for the edits Kara!

place_types.md
kmoscoe: Hmm, then that begs the question of why we are adding them here. It seems that everything on this page is administrative divisions, so maybe we should just omit them after all?  Probably we need Julia to weigh in here.

place_types.md
kmoscoe: Thanks for that info, makes sense. Let's decide first on whether to add the tracts or ot.

place_types.md
pradh: I think the titles could probably be more general than "administrative division"?  Perhaps use the "geographical division"?

https://screenshot.googleplex.com/A7Zjg9ZY8wz3bbb

e.g, I'm not sure that every City is an administrative division

kmoscoe: OK, that makes sense. I'll rename everything to "geographical division". Thanks for the suggestion!

pradh: Alternative, feel free to defer that to Julia, and just add the new types in this PR (given it is not regressing :) )

pradh: I do +1 changing it, please ignore my last comment :)

place_types.md
```diff
@@ -43,17 +43,19 @@ countries, but not all countries.
...
+|[State](https://datacommons.org/browser/State)|U.S. states.<br><br>For example, [California](https://datacommons.org/place/geoId/06) or [Maryland](https://datacommons.org/place/geoId/)|
+|[County](https://datacommons.org/browser/County)|U.S. counties.<br><br>For example, [Santa Clara County](https://datacommons.org/place/geoId/0669084)|
+|[CensusZipCodeTabulationArea](https://datacommons.org/browser/CensusZipCodeTabulationArea)|U.S. zip codes as defined by the U.S. Census Bureau.<br><br>For example, [94043](https://datacommons.org/place/zip/94043).<br><br>While there is significant overlap, these codes don't always correspond to the zip codes used by the US Postal Service.|
+|[CensusBlockGroup](https://datacommons.org/browser/geoId/01003990000)|U.S. block groups as defined by the U.S. Census Bureau.<br><br>For example, [Block Group 0](https://datacommons.org/browser/geoId/010039900000)|
```
pradh, line 57: nit: Tracts contain block groups

pradh (approved):
Thanks for the updates!

place_types.md
kmoscoe: Gaaahh!  Sorry, should be fixed now.

PR #468, opened 2024-08-06
Add info for custom DC in API docs

kmoscoe:


custom_dc/quickstart.md
```diff
@@ -198,13 +198,13 @@ Note that NL support increases the startup time of your server and consumes more
 
 ## Send an API request
 
-A custom instance can accept REST API requests at the endpoint `/core/api/v2/`. To try it out, here's an example request that returns the same data as in the interactive queries above, using the `observation` API. You can enter this query in your browser to get nice output:
+A custom instance can accept [REST API](/api/rest/v2/index.html) requests at the endpoint `/core/api/v2/`. To try it out, here's an example request that returns the same data as in the interactive queries above, using the `observation` API. You can enter this query in your browser to get nice output:
```
keyurva, line 201: nit: perhaps add "here's an example request _you can make to your local Custom DC instance_ that returns..."

keyurva, line 201: > You can enter this query in your browser to get nice output

Can skip this?

keyurva, line 201: We should add that they can use the CDC endpoint for accessing both custom and base dc data.

keyurva (approved):

custom_dc/quickstart.md
kmoscoe: OK

kmoscoe: Great point!  Added.

PR #482, opened 2024-08-21
Fix link and typo

gmechali:
First "Getting started" link in System overview is broken. This cl fixes it by redirecting the user to the getting started page.
Also fixes typo from "wappers" to "wrappers".


kmoscoe (approved):
Thank you for this!!

PR #486, opened 2024-08-22
Remove sections on finding DCID examples etc. and repoint to canonical doc

kmoscoe:


kmoscoe:
Now that we have a "canonical doc" on how to look up DCIDs and find places, let's just point to it.

dwnoble (approved):

PR #491, opened 2024-08-26
Revise procedure to have users just enter name of container directly

kmoscoe:
Per the difficulties Oklahoma had with selecting the correct image, I made it simpler by just having them type in the correct name.


keyurva (approved):
Thanks!

PR #493, opened 2024-09-01
Rewrite procedures for sorting, using workaround

kmoscoe:
The "Create filter view" method no longer works. This PR updates the docs that discuss sorting, to recommend just copying values. It also updates the sleep tutorial to use 2020 (the last year for which we seem to have data) instead of 2016 data.


beets (approved):
thanks for the udpates!

PR #494, opened 2024-09-03
Add instructions for setting up Analytics

kmoscoe:


kmoscoe:
Per a long-standing request from various partners, this PR provides instructions on how to use GA custom events exported by DC to report on things like user search queries.

dwnoble (approved):
Thank you Kara!

PR #498, opened 2024-09-04
Update index.md

clincoln8:


beets (approved):
thanks for the fix!!

for future pr's, please add more context to the pr description so it's easy to get a sense for it when you scan through the pr list

PR #504, opened 2024-09-10
Removed navigation options from docsite header

dwnoble:
- Removed navigation options from docsite header
- Added "Search docs" placeholder to header search
- Added "dot" separator to header title

Before
Screenshot 2024-09-10 at 2 51 59 PM

After
Screenshot 2024-09-10 at 2 51 50 PM


kmoscoe:
The one problem with this approach is that there are some items that are not represented in the left nav, like "FAQ", and some that are but shouldn't be. :-) (like "Papers" and "Blog"). As long as you'll agree to put them in website menus instead, then this is fine.

beets (approved):
we might even try to remove the jquery.min.js and bootstrap.min.js though i'm not sure if there are other pages that now rely on them (e.g. the post/get tabs on our rest api pages). maybe that's for next week.

@kmoscoe we will have those links in the new website headers. this removes the need to keep the two headers in sync..

kmoscoe:
Great, thank you!  👌

PR #507, opened 2024-09-12
Fix incorrect links in tutorials page

kmoscoe:
Some of the links were using the internal "sandbox" host rather than external "research".


chejennifer (approved):
thanks!

PR #512, opened 2024-09-16
Add more detail about data security

kmoscoe:


kmoscoe:
This PR adds more details to the FAQ about data security. It also goes back to the original "Quickstart" name, which is nicer. :-)

custom_dc/faq.md
```diff
@@ -19,9 +19,12 @@ Yes; there are many options for doing so. If you want an entirely private site w
 
 Note that you cannot apply fine-grained access restrictions, such as access to specific data or pages. Access is either all or nothing. If you want to be able to partition off data, you would need to create additional custom instances.
 
-### Will my custom data end up in the base Data Commons?
+### Will my data or queries end up in base Data Commons? {#data-security}
 
-No. Even if a query needs data from the base Data Commons, gleaned from the (incomplete) results from your custom data, it will never transfer any custom data to the base data store.
+Your user queries, observations data, or property values are never transferred to base Data Commons. The NL model built from your custom data lives solely in your custom instance. The custom Data Commons Mixer does make API calls to the base Data Commons Mixer (as depicted in [this diagram](/custom_dc/index.html#system-overview)) only in the following instances:
```
keyurva, line 24: Have we introduced them to mixer anywhere in the document? If not, may be we just say that "The custom Data Commons _instance_ does make API calls to base Data Commons..."

custom_dc/faq.md
```diff
@@ -19,9 +19,12 @@ Yes; there are many options for doing so. If you want an entirely private site w
...
+- At data load time, an API call is made from the custom instance to the base instance to resolve entity names to DCIDs; for example, if your data refers to a particular country name, the custom instance will send an API request to look up its DCID.
```
keyurva, line 25: Again, have we introduced DCID or should we have a blurb about mapping entities to their respective identifiers in DC?

custom_dc/faq.md
```diff
@@ -19,9 +19,12 @@ Yes; there are many options for doing so. If you want an entirely private site w
...
+Your user queries, observations data, or property values are never transferred to base Data Commons. The NL model built from your custom data lives solely in your custom instance. The custom Data Commons Mixer does make API calls to the base Data Commons Mixer (as depicted in [this diagram](/custom_dc/index.html#system-overview)) only in the following instances:
+- At data load time, an API call is made from the custom instance to the base instance to resolve entity names to DCIDs; for example, if your data refers to a particular country name, the custom instance will send an API request to look up its DCID. 
+- At run time, when a user enters an NL query, the custom instance uses its NL model to identify the relevant statistical variables. The custom instance then issues two SQL queries for the statistical variable observations: one to your database and one to the base Data Commons database. These queries only include DCIDs and contain no information about the original query or context of the user request. The data is joined by entity DCIDs.
```
keyurva, line 26: "two SQL queries" is incorrect. There is one sql query to the custom dc's sql instance. The other one is a remote api call to base dc.

custom_dc/faq.md
kmoscoe: Yes, right here: https://screenshot.googleplex.com/BoJEbJzgQZhG4vZ. However, you're right that it's simpler to just say "instance".

custom_dc/faq.md
kmoscoe: Yes, it's mentioned several times in https://docs.datacommons.org/custom_dc/custom_data.html. But I'll add a link here to the glossary entry.

custom_dc/faq.md
kmoscoe: Thanks for the correction:

custom_dc/faq.md
```diff
@@ -21,9 +21,9 @@ Note that you cannot apply fine-grained access restrictions, such as access to s
...
-- At run time, when a user enters an NL query, the custom instance uses its NL model to identify the relevant statistical variables. The custom instance then issues two SQL queries for the statistical variable observations: one to your database and one to the base Data Commons database. These queries only include DCIDs and contain no information about the original query or context of the user request. The data is joined by entity DCIDs.
+Your user queries, observations data, or property values are never transferred to base Data Commons. The NL model built from your custom data lives solely in your custom instance. The custom Data Commons instance does make API calls to the base Data Commons instance (as depicted in [this diagram](/custom_dc/index.html#system-overview)) only in the following instances:
+- At data load time, an API call is made from the custom instance to the base instance to resolve entity names to [DCIDs](/glossary.html#dcid); for example, if your data refers to a particular country name, the custom instance will send an API request to look up its DCID.
```
keyurva, line 25: s/an API call is/API calls are

custom_dc/faq.md
```diff
@@ -21,9 +21,9 @@ Note that you cannot apply fine-grained access restrictions, such as access to s
...
+- At data load time, an API call is made from the custom instance to the base instance to resolve entity names to [DCIDs](/glossary.html#dcid); for example, if your data refers to a particular country name, the custom instance will send an API request to look up its DCID. 
+- At run time, when a user enters an NL query, the custom instance uses its NL model to identify the relevant statistical variables. The custom instance then issues two requests for statistical variable observations: a SQL query to your custom SQL database and an API call to the base Data Commons database. These requests only include DCIDs and contain no information about the original query or context of the user request. The data is joined by entity DCIDs.
 - At run time, when the website frontend renders a data visualization, it will also make the same two SQL queries to get observations data.
```
keyurva, line 27: "same two SQL queries" should be changed to something like "same two requests mentioned above"

premr: USes its "LOCAL" NL model to identify the relevant ...

At Run time - again,  do we want to be explicit that the data itself is not shared with our frontend visualization work - ie all the rendering happens locally?

custom_dc/faq.md
kmoscoe: Fixed.

custom_dc/faq.md
kmoscoe: added "local". 
Re the second comment, I fixed one thing, but I'm not sure what you mean by "shared with frontend visualization work"; I mean I'm guessing there is JS that runs in the browser that a) fetches the data and b) renders the visualizations. What exactly do you want me to say here?

premr:
Ok, yes I think the second statement is now fine. lGTM

PR #527, opened 2024-10-22
Document data docker schema update mode

hqpho:
This mode is added by https://github.com/datacommonsorg/website/pull/4686. A subsequent PR to mixer will link the new docsite page directly from the schema check error message: https://github.com/datacommonsorg/mixer/pull/1440

### DIFF
--- a/custom_dc/database_update.md
+++ b/custom_dc/database_update.md
@@ -23,7 +23,7 @@ This page contains detailed instructions for passing `SCHEMA_UPDATE_ONLY` to the
 
 ## Local data management job with local SQLite database
 
-Add `-e SCHEMA_UPDATE_ONLY=true` to the Docker run command for the data management container (the first command in [this doc section](/custom_dc/custom_data.html#docker-data)):
+Add `-e SCHEMA_UPDATE_ONLY=true` to the Docker run command for the data management container (the first command in [this doc section](/custom_dc/custom_data.html#docker-data){: target="_blank"}):
 
 <pre>
 docker run \
@@ -48,7 +48,7 @@ Run your existing Cloud Run job with an environment variable override.
 
 ## (Advanced) Local data management job with Cloud SQL
 
-If you followed [these instructions](/custom_dc/data_cloud.html#run-local) to load data from your local machine into a Cloud SQL database, add `-e SCHEMA_UPDATE_ONLY=true` to the Docker run command from the final step:
+If you followed [these instructions](/custom_dc/data_cloud.html#run-local){: target="_blank"} to load data from your local machine into a Cloud SQL database, add `-e SCHEMA_UPDATE_ONLY=true` to the Docker run command from the final step:
 
 <pre>
 docker run -it \
--- a/custom_dc/database_update.md
+++ b/custom_dc/database_update.md
@@ -51,18 +51,14 @@ Run your existing Cloud Run job with an environment variable override.
 If you followed [these instructions](/custom_dc/data_cloud.html#run-local){: target="_blank"} to load data from your local machine into a Cloud SQL database, add `-e SCHEMA_UPDATE_ONLY=true` to the Docker run command from the final step:
 
 <pre>
-docker run -it \
+docker run \
 --env-file $PWD/custom_dc/env.list \
--p 8080:8080 \
--e DEBUG=true \
--e GOOGLE_APPLICATION_CREDENTIALS=/gcp/creds.json \
-<b>-e SCHEMA_UPDATE_ONLY=true</b> \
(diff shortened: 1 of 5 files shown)


custom_dc/database_update.md
```diff
@@ -0,0 +1,64 @@
...
+nav_order: 9
+parent: Build your own Data Commons
+---
+
+{:.no_toc}
+# Update your database schema
+
+While starting Data Commons services, you may see an error that starts with `SQL schema check failed`. This means your database schema must be updated for compatibility with the latest Data Commons services.
+
+You can update your database by running a data management job with the environment variable `SCHEMA_UPDATE_ONLY` set to `true`. This will alter your database without modifying already-imported data.
```
keyurva, line 13: May need to be updated based on what we decide in https://github.com/datacommonsorg/website/pull/4686/files#r1815646141

keyurva (approved):
LGTM modulo the decision on the env var.

custom_dc/database_update.md
hqpho: Updated.

hqpho:
Will wait to commit this until the mode is available in the stable data container image.

kmoscoe (commented):
I wouldn't actually include this as a separate page. It's too minor a detail. I would instead just put the text in the relevant existing sections. Can I send you an alternative PR as a proof of concept?

hqpho:
> I wouldn't actually include this as a separate page. It's too minor a detail. I would instead just put the text in the relevant existing sections. Can I send you an alternative PR as a proof of concept?

My thought was to have it on one page for ease of linking from an error message! Is that alright? Maybe we can remove this page from the sidenav if we don't want to feel like we're cluttering the docsite?

kmoscoe:
You could still link to a subsection using an anchor.

kmoscoe:
Another thing I don't really like is repeating in entirety the startup procedures. It makes more sense to integrate this info into the existing procedures. Let me send you a PR so you can see what I mean.

hqpho:
Kara, thanks for putting together an [alternate approach](https://github.com/datacommonsorg/docsite/pull/529)! I understand the desire not to duplicate startup commands across multiple pages. That said, I worry that the other approach:
- Adds cognitive overhead for people running a typical data load command, trying to copy/paste from a doc page and having to learn about a new param that is almost never relevant to them
- Doesn't explain the mechanism by which the startup time is minimized, so it might come as a surprise to people what exactly the mode does (or rather doesn't) do

Another argument I'd make in favor of having schema update mode be its own page is that we may expand the mode variable in the future to support more different modes, in which case we can easily revise and extend that page.

@keyurva Do you want to weigh in as a decision tiebreaker here?

kmoscoe:
On Mon, Oct 28, 2024 at 3:35 PM Hannah Pho ***@***.***> wrote:

> Kara, thanks for putting together an alternate approach
> <https://github.com/datacommonsorg/docsite/pull/529>! I understand the
> desire not to duplicate startup commands across multiple pages. That said,
> I worry that the other approach:
>
>    - Adds cognitive overhead for people running a typical data load
>    command, trying to copy/paste from a doc page and having to learn about a
>    new param that is almost never relevant to them
>
> OK, another possibility is to have a subheading about running in schema
update mode.

>
>    - Doesn't explain the mechanism by which the startup time is
>    minimized, so it might come as a surprise to people what exactly the mode
>    does (or rather doesn't) do
>
> But that wasn't really given in your PR either. I can easily add some more
info if you give it to me.

>
>
> Another argument I'd make in favor of having schema update mode be its own
> page is that we may expand the mode variable in the future to support more
> different modes, in which case we can easily revise and extend that page.
>
I don't like the idea of having pages determined by some random feature.
They should be determined by the general stage in the workflow, which is
the overall structure I've set up and which is linked from the landing
page. Let's please not introduce additional pages for every new feature we
add.

keyurva:
Is there a middle ground where we don't have a separate page for it but are able to make it as self-sufficient as possible?

The typical workflow is likely going to be what Hannah described: a user starts the service, notices the schema failure with a link to the relevant section / page in the docsite and is primarily interested in resolving this failure in the quickest possible manner.

kmoscoe:
Well, the quickest possible way would just say in the error message:
"Restart the data management job, optionally adding the -e
DATA_RUN_MODE=schemaupdate option for faster performance, and rerun the
services job" or something like that.

The middle ground would be a self-contained topic within the existing
pages. Let me prepare a PR for you that would show that as a PoC, OK?

kmoscoe:
Hey both -- I sent you PR 530
<https://github.com/datacommonsorg/docsite/pull/530> to try to make the
text more standalone. Please review; thanks!

hqpho:
Subsumed by https://github.com/datacommonsorg/docsite/pull/530

PR #532, opened 2024-11-04
Address feedback from one.org et. al.

kmoscoe:


kmoscoe:
This PR addresses all the feedback from one.org, pending the automatic triggering of job restarts:
1) Adds command-line equivalents for all repetitive procedures
2) Adds a section on storing secrets in Secret Manager and updates screenshots for that
3) Removes incorrect cost estimate

In a future PR, will convert the CLI/Console procedures to tab-based.

dwnoble (approved):

PR #536, opened 2024-11-07
New procedure for syncing to stable release

kmoscoe:
Better Github procedure suggestions welcome!


custom_dc/build_image.md
```diff
@@ -41,82 +41,79 @@ If you want to pick up the latest prebuilt version, do the following:
...
+   ```
+   Rather than developing on this default branch, we recommend that you create another branch, as described in the next step.
+
+#### Create a remote and sync a local branch to the stable branch
+This procedure assumes that you are using Github and have already created a forked repo of `datacommonsorg/website`.
 
+1. From the `website` or other directory where you have cloned the code, create a remote that points to your fork, and only uses the stable branch:
    <pre>
-   git checkout -b <var>BRANCH_NAME</var> customdc_stable
+   git remote add -t customdc_stable <var>REMOTE_NAME</var> <var>FORK_URL</var>
```
hqpho, line 82: I think we should probably have this track `customdc_stable` on the main repo rather than the fork, so the user doesn't have to keep their fork in sync. WDYT?

kmoscoe: You're right, that makes more sense! So in that case, the fork is actually kind of irrelevant here, because they will only push to it, rather than pull from it, right?

custom_dc/build_image.md
```diff
@@ -41,82 +41,79 @@ If you want to pick up the latest prebuilt version, do the following:
...
    <pre>
-   git checkout -b <var>BRANCH_NAME</var> customdc_stable
+   git remote add -t customdc_stable <var>REMOTE_NAME</var> https://github.com/datacommonsorg/website
    </pre>
-
-1. To verify that your local repo is at the same version of the code, run the following command:
-
+1. Create a new branch synced to the stable branch:
+   <pre>
+   git checkout -b <var>BRANCH_NAME</var> origin/customdc_stable
```
hqpho, line 86: Since the remote name is hard-coded here, how about we recommend something for it both here (line 86) and above (line 82)? I'd recommend calling it `upstream` for consistency with our other docs and tools.

kmoscoe: Yeah, that's a good point, I do what you suggest, which is that origin is my fork and upstream is the original. Let me rewrite this accordingly -- including mentioning that we assume that they keep the default fork name of "origin", in which case this part should actually their remote name, right?

kmoscoe: Actually, no I just realized that this is the opposite of what you get by default when you run git clone.

kmoscoe: So in reality, they don't even need to this step at all!

hqpho:
Thanks for putting this together! I think probably once we get feedback from partners on how exactly they are using the branch, we might want to update this to also include how to pull in new changes from the branch. And we might want to recommend an approach that involves a fork by default. But for now this is good enough to get people started!

kmoscoe:
OK, try reviewing again now.

custom_dc/build_image.md
```diff
@@ -41,84 +41,68 @@ If you want to pick up the latest prebuilt version, do the following:
...
-
-> **Note:** If you are working on a large-scale customization, we recommend that you use a version control system to manage your code. We provide procedures for Github, and assume the following:
-- You have a Github account and project.
-- You have created a fork off the base Data Commons `website` repo (https://github.com/datacommonsorg/website){: target="_blank"} and a remote that points to it, and that you will push to that fork. 
-
-
-### Sync a local workspace to the stable release
+Rather than building from the master branch, which includes the very latest changes in Github, that may not have been tested, we recommend that you use the tested "stable" branch equivalent of the stable Docker image. This branch is `customdc_stable`, and is available at [https://github.com/datacommonsorg/website/tree/customdc_stable](https://github.com/datacommonsorg/website/tree/customdc_stable){: target="_blank"}.
```
hqpho, line 48: Optional minor grammar thing: "that may not have been tested" kind of makes this sentence not flow IMO. WDYT of splitting into two sentence? "Building from the master branch includes the very latest changes in Github, which may not have been tested. Instead, we recommend that you use..."

custom_dc/build_image.md
```diff
@@ -41,84 +41,68 @@ If you want to pick up the latest prebuilt version, do the following:
...
 
-If you are using a version control system other than Github, you can download a ZIP or TAR file from [https://github.com/datacommonsorg/website/releases/tag/customdc_stable](https://github.com/datacommonsorg/website/releases/tag/customdc_stable){: target="_blank"}. 
+> **Note:** If you are working on a large-scale customization, we recommend that you use a version control system to manage your code. We provide some procedures for Github.
 
-In Github, use the following procedure.
+#### Clone the stable branch only
```
hqpho, line 52: Nit: Skips a header level from the previous header; could maybe change to `###`?

custom_dc/build_image.md
```diff
@@ -41,84 +41,68 @@ If you want to pick up the latest prebuilt version, do the following:
...
-   git checkout -b <var>BRANCH_NAME</var> customdc_stable
+   cd website | cd <var>DIRECTORY</var>
    </pre>
 
-1. To verify that your local repo is at the same version of the code, run the following command:
-
+1. Update files:
    ```
-   git log --oneline --graph
+   git pull origin customdc_stable
```
hqpho, line 84: Can you confirm that since they have only cloned the stable branch there's no need to do `git checkout origin/customdc_stable`?

custom_dc/build_image.md
```diff
@@ -41,84 +41,68 @@ If you want to pick up the latest prebuilt version, do the following:
...
    ```
-   You should see output similar to the following:
+   Note that`origin` here refers to the source `datacommonsorg/website` repo. You may be using another remote name to point to that repo.
```
hqpho, line 86: Nit: missing space before `origin`

custom_dc/build_image.md
kmoscoe: Good catch, I missed that.

custom_dc/build_image.md
kmoscoe: So this is the part we took out -- which is creating a new branch and working off that. If they use your command, they get into detached HEAD mode, which they don't want. What they want is git checkout -b <newbranchname> origin/customdc_stable, but we agreed to remove that step? (I'm happy to put it back in :-))

custom_dc/build_image.md
kmoscoe: Fixed, thanks for catching.

custom_dc/build_image.md
hqpho: This works as-is! I just tested and since they only clone the one branch, there's no need to also check out the branch (it's checked out by default).

hqpho (approved):

PR #540, opened 2024-11-12
New navigation

kmoscoe:
This PR implements the navigation proposed at https://docs.google.com/document/d/1tp2hLKSJXgSFJcED3N40WCMbEOoj8MfZy4mtj6JvsWM/edit?resourcekey=0-OX5FrCW7cb0RN3xKUsg3sg&, using the existing front-matter and sidebar processing methods.  In particular, it makes the following changes:

- The "how to use" page becomes the default landing page
- The standalone pages are now under this top-level heading
-  Sheets and Web Components are pulled out from under "API" and become top-level headings
- All headings use a consistent task-oriented style
- Subpages of datasets are removed from nav
- Redundant headings (e.g. "Blog") are removed from nav
- "Import datasets", which can in no way be used by clients, is unpublished now

Still to be handled: "Courseware".


papers/DataGemma-FullPaper.pdf
beets, line 1: could we keep these pdf's (even if we don't link to them ourselves, there might be straggler links around)

beets (approved):
sorry for the late review on this! just one comment on keeping the pdf of the papers here, but otherwise lgtm :)

papers/DataGemma-FullPaper.pdf
kmoscoe: Really? OK.

PR #543, opened 2024-11-18
Add tabs for GCP procedures

kmoscoe:
This PR implements tab selectors for console/CLI procedures (and will add Terraform in the future) and does some cleanup:

- Adds a new JS file for allowing multiple tab groups on a single page (currently not supported by API doc JS)
- Adds new CSS definitions for the tabs, for styling similar to Google Cloud docs
- Removes duplicate CSS in custom.scss and styles.css (these are only used by the REST API v0 docs)
- Cleans up some pre formatting 
- Changes the code font to match the pre one


dwnoble (approved):
Thanks Kara!

PR #548, opened 2024-11-26
Fixed redirect from /tutorials to /api/python/tutorials

dwnoble:
Fixed redirects:

* `/tutorials` -> `/api/python/tutorials`
*  `/bigquery/data_in_bq` ->
* `/bigquery/unique_identifiers` -> `/data_model.html`
* `/bigquery/unique_identifiers#dcid` -> `/data_model.html` (the `#` is ignored)

### DIFF
--- a/api/python/index.md
+++ b/api/python/index.md
@@ -5,7 +5,7 @@ nav_order: 30
 parent: API - Query data programmatically
 has_children: true
 redirect_from:
-    /tutorials/
+    /tutorials
     /tutorials/index
 ---
 
--- a/api/python/index.md
+++ b/api/python/index.md
@@ -4,9 +4,6 @@ title: Python
 nav_order: 30
 parent: API - Query data programmatically
 has_children: true
-redirect_from:
-    /tutorials
-    /tutorials/index
 ---
 
 # Data Commons Python API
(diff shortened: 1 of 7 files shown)


kmoscoe (commented):
You can remove the tutorials/index.md. The redirect_from makes it unnecessary.

dwnoble:
When i remove tutorials/index.md i get a 404

api/python/index.md
```diff
@@ -5,7 +5,7 @@ nav_order: 30
 parent: API - Query data programmatically
 has_children: true
 redirect_from:
-    /tutorials/
+    /tutorials
```
kmoscoe, line 8: This is actually the wrong file. It should be in api/python/tutorials.md. AND you need hyphens before the path if there are multiple entries.

tutorials/index.md
```diff
@@ -0,0 +1,4 @@
+---
```
kmoscoe, line 1: You can revert this; this should not be needed.

api/python/index.md
dwnoble: Fixed- this now redirects to `/api/python/tutorials`

tutorials/index.md
dwnoble: Removed

clincoln8 (approved):
Thanks Dan!

kmoscoe (approved):

PR #562, opened 2025-01-14
Move the Feedback section from index page to FAQ

kmoscoe:


keyurva (approved):
Thank you for the super quick fix!

PR #564, opened 2025-01-16
Clarified usage of resolve API

kmoscoe:
Clarified that Resolve API cannot be used to get DCIDs of linked nodes.
Staged at bullie.svl.corp.google:4000


chejennifer (approved):
thanks!

PR #566, opened 2025-01-27
Improve per-variable property description for custom DC

kmoscoe:
In response to Tech Soup's question, this gives more information about using measuredProperty to control display of variables when using "implicit" schema.

Also makes output directory optional in Docker commands.


custom_dc/build_image.md
```diff
@@ -33,7 +33,7 @@ If you want to pick up the latest prebuilt version, do the following:
    <pre>docker run \
    --env-file $PWD/custom_dc/env.list \
    -v <var>INPUT_DIRECTORY</var>:<var>INPUT_DIRECTORY</var> \
-   -v <var>OUTPUT_DIRECTORY</var>:<var>OUTPUT_DIRECTORY</var> \
+   [-v <var>OUTPUT_DIRECTORY</var>:<var>OUTPUT_DIRECTORY</var>] \
```
keyurva, line 36: I think we had decided to keep it in the docs to be more explicit and avoid errors?

keyurva (approved):
measuredProperty bit LGTM. Not sure about removing output directory

kmoscoe:
OK, I'll revert those.

PR #567, opened 2025-01-28
Fix observation properties to not confusingly refer to variables!

kmoscoe:


keyurva (approved):
Thanks for the catch and fix!

PR #577, opened 2025-03-04
Remove references on contributing page to removed form

kmoscoe:
Nobody was actually using the "contributing in other ways" form, and we shouldn't use Google Forms to collect email addresses. So removing all references  to the catchall form.


miss-o-soup (approved):

n-h-diaz (approved):
Thanks!

PR #581, opened 2025-03-07
Add new jekyll workflow

beets:
per instructions here: https://jekyllrb.com/docs/continuous-integration/github-actions/

### DIFF
--- a/.github/workflows/jekyll.yml
+++ b/.github/workflows/jekyll.yml
@@ -1,0 +1,65 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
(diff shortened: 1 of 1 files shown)


clincoln8 (approved):

.github/workflows/jekyll.yml
```diff
@@ -0,0 +1,65 @@
...
+# Sample workflow for building and deploying a Jekyll site to GitHub Pages
+name: Deploy Jekyll site to Pages
+
+on:
+  # Runs on pushes targeting the default branch
+  push:
+    branches: ["master"]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
```
clincoln8, line 15: this looks empty, is that intentional?

beets: thanks for the catch - commenting it out

PR #585, opened 2025-03-07
Added content for observation filters in REST API

kmoscoe:
Added descriptions and usage examples for the two filters available for the observation endpoint, per Jorge's request.


n-h-diaz (approved):
Thanks!

PR #586, opened 2025-03-10
Remove content from datasets pages and point to website

kmoscoe:
Add redirects to website from all pages


miss-o-soup:
Hi Kara,

For the redirects, maybe worth point the Category page to the same category tab. 
For instance for Agriculture, we could re-direct to:
https://datacommons.org/data/agriculture
vs just
https://datacommons.org/data

What do you think?

datasets/Agriculture.md
miss-o-soup, line 1: Hi Kara,

For the redirects, maybe worth point the Category page to the same category tab. 
For instance for Agriculture, we could re-direct to:
https://datacommons.org/data/agriculture
vs just
https://datacommons.org/data

What do you think?

datasets/Crime.md
miss-o-soup, line 1: Same thing, the one here could be: 
https://datacommons.org/data/crime

kmoscoe: Yes, I did think about that, but I was concerned that pages might move around or disappear, so wanted to use a link I wouldn't have to update again. :-)  But yeah, I guess I should do that. I'm going to remove the biomedical link though.

miss-o-soup (approved):
Approved - thank you!

PR #588, opened 2025-03-14
New node page for Python V2 client

kmoscoe:


api/python/v2/node.md
```diff
@@ -0,0 +1,586 @@
...
+}
+```
+{: .response-signature .scroll}
+
+#### Example 3: Get a list of all statistical variables
+
+This example gets the list of all statistical variables in the knowledge graph, by fetching all nodes that are types of the class `StatisticalVariable` and using the `<-typeOf` symbol to express the incoming relationships. Also, because of the size of the response, it enables [pagination](#pagination) to split up the response data into multiple calls.
+
+```python
+>>> >>> print(client.node.fetch(node_dcids=["StatisticalVariable"], expression="<-typeOf", all_pages=False).to_json())
```
dwnoble, line 152: Looks like there's an extra `>>>` here

api/python/v2/node.md
```diff
@@ -0,0 +1,586 @@
...
+
+### Examples
+
+{: #fetch_ex1}
+#### Example 1: Get all incoming property labels for a given node
+
+This examples gets all incoming arc property labels, i.e. the property labels of attached nodes, for the node with DCID `geoId/06` (California) by querying with the `<-` symbol. This returns just the property labels but not the property values. 
+
+```python
+>>> print(client.node.fetch(node_dcids=["geoId/06"], expression="<-").to_json())
```
dwnoble, line 105: I dont think the `>>>` is necessary here. I'd also remove the `print` and `to_json()` calls- they are a little distracting from the core library.

An alternative way to present it could be two blocks like:

This example gets all incoming arc....
```
client.node.fetch(node_dcids=["geoId/06"], expression="<-")
```

This returns just the property labels but not the property values. 
```
{
  "data": {
    "geoId/06": {
      "properties": [
        "affectedPlace",
        "containedInPlace",
        "location",
        "member",
        "overlapsWith"
      ]
    }
  }
```

And just explain the `to_json()` helper separately

api/python/v2/node.md
```diff
@@ -0,0 +1,586 @@
...
+
+| Name          | Type  |   Description  |
+|---------------|-------|----------------|
+| node_dcids <br/> <required-tag>Required</required-tag> | string or list of strings | List of the [DCIDs](/glossary.html#dcid) of the nodes to query.  |
+| expression <br/> <required-tag>Required</required-tag> | string  | Property to query, represented with symbols including arrow notation. For more details, see [relation expressions](/api/rest/v2/#relation-expressions). By using different relations, you can query node information in different ways, such as getting the edges and neighboring node values. Examples below show how to request this information for one or multiple nodes. |
+| all_pages <br/> <optional-tag>Optional</optional-tag> | bool | Whether all data should be sent in the response. Defaults to `True`. Set to `False` to return paginated responses. See [Pagination](#pagination) for details. |
+| next_token <br/> <optional-tag>Optional</optional-tag> | string | If `all_pages` is set to `False`, set this to the next token returned by the previous response. Defaults to `None`. See [Pagination](#pagination) for details. |
+{: .doc-table }
+
+### Examples
```
dwnoble, line 97: Before getting into the examples, could you add a section about how to import & initialize the library?

something like:
```
import datacommons

client = datacommons.Client(api_key="<YOUR_KEY>")
```

api/python/v2/node.md
```diff
@@ -0,0 +1,586 @@
...
+
+#### Example 1: Get one (outgoing) property value for a given node
+
+This example gets the `name` property for a given node with DCID `dc/03lw9rhpendw5`. This is identical to [example 2](#fetch_ex2) of the `fetch` method.
+
+```python
+>>> {
```
dwnoble, line 279: I think this is missing the call to fetch_property_values

dwnoble (approved):

api/python/v2/node.md
kmoscoe: Fixed. thanks.

api/python/v2/node.md
kmoscoe: Ahh, my bad, I didn't add you to a reviewer on the previous PR, https://github.com/datacommonsorg/docsite/pull/576, where it lives. Added you now!

api/python/v2/node.md
kmoscoe: Fixed

api/python/v2/node.md
dwnoble: Lgtm! thanks!

dwnoble (approved):
Thanks Kara!

kmoscoe:
Jorge -- I'm gonna merge this, because I'm working on so many branches at a time, I can't keep track of what I've done. :-) Please review it when you get a chance.

PR #589, opened 2025-03-20
Add doc about facets without observations, add 3 examples and do some cleanup

kmoscoe:
This adds 3more examples to the REST observations page, adds information about select=facet, with an example, and cleans up some other pages (adds TOCs etc.)


api/python/v2/node.md
dwnoble, line 1: Was this deletion intentional?

dwnoble (approved):

api/python/v2/node.md
kmoscoe: No, it wasn't, I need to fix this somehow.

kmoscoe: Fixed now, thanks to Christie.

PR #599, opened 2025-03-30
Rename a method and parameter, fix some incorrect parameters, and add links to source code

kmoscoe:


api/python/v2/resolve.md
```diff
@@ -287,21 +289,21 @@ Response:
 ```
 {: .example-box-content .scroll}
 
-## fetch_dcid_by_wikidata_id
+## fetch_dcid_by_wikidata_ids
```
clincoln8, line 292: should this be fetch_dcids_by_wikidata_id (add "s" after dcid and remove "s" after wikidata_id)?

https://github.com/datacommonsorg/api-python/blob/master/datacommons_client/endpoints/resolve.py#L84

api/python/v2/resolve.md
```diff
@@ -316,7 +318,7 @@ Request:
 {: .example-box-title}
 
 ```python
-resolve.fetch_dcid_by_wikidata_id(wikidata_id="Q30")
+resolve.fetch_dcid_by_wikidata_ids(wikidata_ids="Q30")
```
clincoln8, line 321: should the "s" be after dcid instead of wikidata_id in the method name?

api/python/v2/resolve.md
kmoscoe: Yes!  OMG, I can't believe I screwed this up!   Thank you so much for catching it!!!

clincoln8 (approved):
thanks Kara!

PR #604, opened 2025-04-09
Update all custom DC docs to use automation script

kmoscoe:
This PR updates all Docker commands to give the alternative (easier) way of doing everything using a Bash script.


dwnoble (approved):
Thanks Kara, this is much simpler

PR #616, opened 2025-05-06
Add links for new tutorials

kmoscoe:


api/python/v2/tutorials.md
```diff
@@ -18,6 +18,8 @@ Example [Google Colab
 notebooks](https://colab.research.google.com/notebooks/intro.ipynb) written in
 Python:
 
+-  [Analyzing U.S. Census Data with Data Commons](https://github.com/datacommonsorg/api-python/blob/master/notebooks/v2/analyzing_census_data.ipynb){: target="_blank"})
```
clincoln8, line 21:
```suggestion
-  [Analyzing U.S. Census Data with Data Commons](https://github.com/datacommonsorg/api-python/blob/master/notebooks/v2/Analyzing_Census_Data_with_Data_Commons.ipynb){: target="_blank"})
```

clincoln8 (approved):

api/python/v2/tutorials.md
kmoscoe: Thanks for catching this!. We now have an inconsistent file naming scheme for the notebooks. Do you think it should be consistent, in which case I will change the file name?

clincoln8: sure, this is probably the better solution -- thanks!!

PR #617, opened 2025-05-07
Python

kmoscoe:


clincoln8 (approved):
Thanks Kara! This might change based on how we decide to handle fetch_place_children vs descendants, but lgtm for now :)

PR #618, opened 2025-05-07
Add examples of new response methods (remove TODOs)

kmoscoe:


kmoscoe:
This PR adds examples for the special response methods for observation and resolve endpoints. Also corrects a few errors.

clincoln8 (approved):
Thanks for the updates!

PR #634, opened 2025-07-09
Remove all references to v1 tutorials and just point to v2

kmoscoe:


courseware/index.md
```diff
@@ -27,7 +27,3 @@ A general audience course focused on the basics of data comprehension, interpret
 **[Intro to Data Science](/courseware/intro_data_science.html)**
 
 An introduction to key data science concepts, targeted at an undergraduate audience. Assumes basic python programming skills.
-
-## Feedback
```
clincoln8, line 31: Is this an intentional deletion? If so, it might be helpful to make a note in the PR description to document rationale for deleting.

clincoln8 (approved):
Thanks!

PR #643, opened 2025-09-30
Clarify that groupStatVarsByProperty should only be used for variablePerRow

kmoscoe:
Also fix some broken links


custom_dc/config.md
```diff
@@ -265,19 +266,19 @@ group
...
 
-## groupStatVarsByProperty
+## groupStatVarsByProperty (explicit schema only)
 
 Optional: Causes the Statistical Variable Explorer to create a top-level category called "Custom Variables", and groups together variables with the same population types and measured properties. For example:
 
 ![group_screenshot](/assets/images/custom_dc/customdc_screenshot10.png){: width="400"}
 
-For explicit schema (which does not give you a `group` option in the `config.json`), if you would like your custom variables to be displayed together, rather than spread among existing categories, this option is recommended.
+For explicit schema, if you would like your custom variables to be displayed together, rather than spread among existing categories, this option is recommended.
```
dwnoble, line 281: This sentence reads a little awkwardly to me

kmoscoe: How so? Can you suggest an improvement?

dwnoble: Maybe: "For explicit schemas, use this option to group your custom variables together"

dwnoble (approved):
Thanks Kara! I don't feel strongly about my comment above, so ill defer to you on it.

custom_dc/config.md
kmoscoe: I like it!  My only concern is that I think this shouldn't actually be used for grouping them together, but rather for grouping them by property. But since the former option doesn't actually exist, your text is much clearer, thank you!

kmoscoe: Actually, you know what? The previous text actually says it all, so this sentence is actually redundant. (Before I had the specification that it is for variablePerRow only, this sentence was needed to contrast the different behaviors. Now that it's clear, this sentence isn't needed at all.)

PR #683, opened 2026-02-25
Add best practices for statvars for Custom DC

kmoscoe:
This PR makes the following changes:
- Describes when to reuse existing statvars vs. creating new ones
- Adds naming convention guidelines (based on go/dcguidebook)
- Fixes description of groupStatVarsByProperty to cover actual behavior
- Expands statvar groups section to include existing variables

Next PR will cover similar content for custom entities and enums


custom_dc/custom_data.md
```diff
@@ -58,7 +58,10 @@ In addition, even if you aggregate by geographical area, you may want to measure
...
+Schema.org and the base Data Commons knowledge graph define entity types for just about everything in the world. An _entity type_ is a high-level concept, and is derived directly from a [`Class`](https://datacommons.org/browser/Class){: target="_blank"} type. Non-place entities are of two types:
+- The thing you are measuring, known as the `populationType` in Data Commons. Often this is a `Person`, which is a commonly used population in Data Commons. But it could be something else entirely, like the beds in a hospital, the price of a commodity, Olympic medals won by a country, or the surface area of an ocean.
```
beets: would be nice if we linked to the pop types for these examples

beets (approved):
wow, this is a great update. thanks for breaking down all the requirements.

there is currently a limit to the length of dcid's in custom dc. i'll ping that group to see where we are with it

might want to wait for ajai to approve this pr too

custom_dc/config.md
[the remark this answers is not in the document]
kmoscoe: Unfortunately, no. The config.json doesn't have any way to reference MCF files, for entities or variables.

custom_dc/custom_data.md
beets: yes, we could link to the "Person" browser page. i'm also wondering if we have poptypes for the others in the examples (beds in a hospital, price of a commodity, etc). in other words, i thought the reader might be curious about how those examples would be modeled.

kmoscoe: The thing is it's a bit confusing: the incoming populationType arcs section shows statvars (https://screenshot.googleplex.com/ASWnrHqVefUfW5r) while it's the domainIncludes section that actually shows the population types: https://screenshot.googleplex.com/BrTds7ghnFBvekP. I think at this point in the doc it's too early to get into these details, which are actually quite confusing.

beets (approved):

PR #694, opened 2026-03-31
Add fields missing from facets in observation response 

kmoscoe:
This PR corrects an oversight that neglected to include 3 extra optional fields in the facet returned in the observation API response. It also changes one of the examples (in both REST and Python) to a different kind of statvar that shows units.


api/python/v2/observation.md
```diff
@@ -65,10 +65,13 @@ With `select=["date", "entity", "variable", "value"]` in effect (the default), t
     }
   "facets" {
     "<var>FACET_ID</var>": {
-      "importName": "...",
-      "provenanceUrl": "...",
-      "measurementMethod": "...",
-      "observationPeriod": "..."
+      "importName": "<var>DATASET_NAME</var>",
```
n-h-diaz: (see other comment) not sure if we're explicitly distinguishing between provenance and dataset

n-h-diaz (approved):
Thanks for the update!

api/rest/v2/observation.md
```diff
@@ -241,7 +253,14 @@ With`select=variable` and `select=entity` only, the response looks like the foll
 |-------------|--------|-------------------------------------|
 | orderedFacets | list of objects | Metadata about the observations returned, keyed first by variable, and then by entity, such as the date range, the number of observations included in the facet etc. |
 | observations | list of objects | Date and value pairs for the observations made in the time period |
-| facets | object | Various properties of reported facets, where available, including the provenance of the data, etc. |
+| facets | object | Various properties of reported facets, where available. |
+| facets.importName | string | The name of the [provenance](/data_model.html#sources) or [dataset](/data_model.html#sources). |
```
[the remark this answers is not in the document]
kmoscoe: You're right that it is technically the "provenance" name. However, in many (most?) the provenance and dataset are the same. Since "provenance" is often misunderstood, I used "dataset" as a more intuitive name. But if you think it's bad to make that shortcut, I can change it.

api/python/v2/observation.md
kmoscoe: Yeah, see my above comment. Provenance is such a confusing concept that I hate to use it. But if you think the inaccuracy is bad, I will fix it.

PR #696, opened 2026-04-08
Fix some Docker commands to use a variable instead of hard-coded 'custom' directory

kmoscoe:
AS a follow-up to PR #693, this fixes some Docker commands that were using 'custom' directory in file paths. It also makes some other updates to custom data docs to make the input directory more clear.


nick-nlb (approved):
LGTM, with the only required change before merging the typo, and the otherone optional.

PR #708, opened 2026-06-10
Add llms.txt and llms-full.txt

kmoscoe:
Here is my curated llms.txt and auto-generated llms-full.txt (done with llms_txt2ctx). I will test them with the list of queries in my proposal.


rohitkumarbhagat (approved):
Thanks Kara.

PR #718, opened 2026-07-06
Remove nav entry for Python V1 API and redirect

kmoscoe:


n-h-diaz (approved):
Thanks!

would using redirect_from + relative path work for this?

kmoscoe:
> Referen
Oh god, of course, that's the best way to do it. Thanks for catching that!

PR #719, opened 2026-07-14
Rewrite statvar content to conform to new guidelines

kmoscoe:
The custom entities file will come in a separate PR.

Staged at http://bullie.svl.corp.google.com:4000/custom_dc/custom_data.html

Please ignore all the formatting changes; my editor decided it was time to comply with some standard I never heard of.


custom_dc/custom_data.md
```diff
@@ -176,66 +176,107 @@ Nodes in the Data Commons knowledge graph are defined in Metadata Content Format
...
+| Indicator | Year | Country code | Country | Sex | Value |
+|-----------|------|--------------|---------|-----|-------|
+| Prevalence of current cigarette smoking among adolescents	| 2018	| ARG	| Argentina	| Male	| 15.5 |
+| Prevalence of current cigarette smoking among adolescents	| 2018	| ARG	| Argentina	| Female | 20 |
+| Prevalence of current cigarette smoking among adolescents	| 2018	| ARG	| Argentina	| Both sexes	| 18 |
+| Prevalence of current cigarette smoking among adolescents	| 2022	| ARM	| Armenia	| Male	| 17 |
+| Prevalence of current cigarette smoking among adolescents	| 2022	| ARM	| Armenia	| Female	| 0 |
+| Prevalence of current cigarette smoking among adolescents	| 2024	| AUT	| Austria	|	Male	| 17.2 |
+| Prevalence of current cigarette smoking among adolescents	2024	| AUT	| Austria	|	Female	| 21.3 |
+| Prevalence of current cigarette smoking among adolescents	2024	| AUT	| Austria	| Both sexes	| 19.5 |
```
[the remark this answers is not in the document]
kmoscoe: wrong

custom_dc/custom_data.md
```diff
@@ -198,43 +198,38 @@ The first thing to notice is that there is a breakdown by sex. Therefore, we'll
...
+* Adolescence is represented by [`Adolescents`](https://datacommons.org/browser/Adolescents){: target="_blank"}, which is a member of the [`AgeGroupClassificationEnum`](https://datacommons.org/browser/AgeGroupClassificationEnum){: target="_blank"}.
 * Sex is represented by an existing property, [gender](https://datacommons.org/browser/gender){: target="_blank"}, which is of type [`GenderType`](https://datacommons.org/browser/GenderType){: target="_blank"}, an enumeration. Its members [Female](https://datacommons.org/browser/Female){: target="_blank"} and [Male](https://datacommons.org/browser/Male){: target="_blank"} already exist as well.
 * [Smoking](https://datacommons.org/browser/Smoking){: target="_blank"}, which is a member of the enumeration [HealthBehaviorEnum](https://datacommons.org/browser/HealthBehaviorEnum){: target="_blank"}.
 
 The MCF would therefore look like this:
 
 ```
-Node: dcid:who/Ratio_Smokers_Age13to17
+Node: dcid:who/Ratio_Smokers_Adolescents
```
carojahn, line 208: what is who/ doing in the definition?

custom_dc/custom_data.md
```diff
@@ -198,43 +198,38 @@ The first thing to notice is that there is a breakdown by sex. Therefore, we'll
...
 typeOf: schema:StatisticalVariable
 name: "Prevalence of current cigarette smoking among adolescents (%)"
 populationType: schema:Person
 measuredProperty: dcid:Smoking
 statType: dcid:Ratio
```
carojahn, line 214: is this a ratio or a rate?

custom_dc/custom_data.md
```diff
@@ -198,43 +198,38 @@ The first thing to notice is that there is a breakdown by sex. Therefore, we'll
...
-constraintProperties: dcid:age
-age: dcid:Years13to17
```
carojahn: Given you are dividing it by a statvar that has age as a constraintProperty (vs adolescent as populationType), one may ask why you are doing it one way on one statvar and differently on another? Like why not have both be person as the populationType and age as constraintProperty? if you know the denominator is 13-17, don't you also know that about the numerator? Or the dataset doesn't specific?

custom_dc/custom_data.md
carojahn: nitpick but if dcid is a strict replication of the values in property it should be Ratio_Smoking_Adolescent

custom_dc/custom_data.md
```diff
@@ -198,43 +198,38 @@ The first thing to notice is that there is a breakdown by sex. Therefore, we'll
...
 * [Smoking](https://datacommons.org/browser/Smoking){: target="_blank"}, which is a member of the enumeration [HealthBehaviorEnum](https://datacommons.org/browser/HealthBehaviorEnum){: target="_blank"}.
 
 The MCF would therefore look like this:
 
 ```
-Node: dcid:who/Ratio_Smokers_Age13to17
+Node: dcid:who/Ratio_Smokers_Adolescents
 typeOf: schema:StatisticalVariable
 name: "Prevalence of current cigarette smoking among adolescents (%)"
 populationType: schema:Person
```
carojahn: if you are deleting age constraintProperty, this should be Adolescent. FWIW I think we should keep Person and keep constraintProperty for age

carojahn (changes requested):

custom_dc/custom_data.md
kmoscoe: WHO. The prefix is explained below.

kmoscoe: > nitpick but if dcid is a strict replication of the values in property it should be Ratio_Smoking_Adolescent

Yeah, but I think we can add a little leeway just to make it a bit more English-y, no?

custom_dc/custom_data.md
kmoscoe: Gah, good catch, I forgot to change that. Fixed now.

custom_dc/custom_data.md
kmoscoe: Well, since we don't have rate defined as a node yet (we should add it), and technically it really is a ratio, I think it's better to keep it like this.

custom_dc/custom_data.md
kmoscoe: Sorry, that was a mistake. I fixed it now: the population type should be Adolescents.

custom_dc/custom_data.md
carojahn: Should your property value be more English-y in that case?

custom_dc/custom_data.md
carojahn: I still struggle with ratio vs rate - where I assumed ratio is more of a "3:1 part" type of expression. But makes sense for now.

custom_dc/custom_data.md
```diff
@@ -207,31 +207,34 @@ The MCF would therefore look like this:
...
 populationType: dcid:Adolescents
 measuredProperty: dcid:Smoking
 statType: dcid:Ratio
 measurementDenominator: dcid:Count_Person_13To17Years
 
 Node: dcid:who/Ratio_Smokers_Adolescents_Female
 typeOf: schema:StatisticalVariable
-name: "Prevalence of current cigarette smoking among adolescents (%) [Female]"
+name: "Prevalence of current cigarette smoking among adolescents (female)"
+description: "Percentage of adolescents who are current cigarette smokers and are female"
```
carojahn: Not strongly opinionated about it: If we are saying that Percentage is a unit - is this really the percentage or the proportion of? 

Nitpick: the description right now feels like a different word order of the name vs adding more value.

custom_dc/custom_data.md
```diff
@@ -207,31 +207,34 @@ The MCF would therefore look like this:
...
 populationType: schema:Adolescents
 measuredProperty: dcid:Smoking
 statType: dcs:Ratio
 constraintProperties: dcid:gender
 gender: dcid:Female
-measurementDenominator: dcid:Count_Person_13To17Years_Female
+measurementDenominator: dcid:Count_Person_13To17Years
```
carojahn: Why are you dividing by all adolescent? Vs by all female adolescent?

custom_dc/custom_data.md
```diff
@@ -261,7 +264,7 @@ All fields that do not reference another node must be in quotation marks.
 
 The following fields are optional:
```
carojahn, line 265: Is this meant to list all StatVar optional fields? ie. should comparisonPeriod be here? domainIncludes, rangeIncludes?

custom_dc/custom_data.md
```diff
@@ -261,7 +264,7 @@ All fields that do not reference another node must be in quotation marks.
...
 
-* `description`: A more detailed textual description of the variable.
+* `description`: A more detailed textual description of the variable. Although this is optional, it is highly recommended, as the description is used to generate embeddings for natural-language search.
 * `measurementQualifier`: This is used to qualify the measurement represented in all observations using the variable. It should be a member of an enumeration, e.g. `Weekly`, `Monthly`, `Annual`, which are members of the [StatAccumulationPeriodEnum](https://datacommons.org/browser/StatAccumulationPeriodEnum){: target="_blank"} type. For instance, if the `measuredProperty` is income, you can use `Annual` or `Monthly` to distinguish income over different periods. If the time interval affects the meaning of variable and and values change significantly by the time period, you should use this field keep them separate.
```
carojahn: Can we add a bit more detail on this one? Assuming it's not defined elsewhere.

Raw notes:
"How was data treated or adjusted?" 
Applies operational filters or scales.

Do’s:
Macroeconomics: Nominal, RealValue, or PurchasingPowerParity (PPP).
Time-series aggregation: Annual, Monthly, or Weekly (used when an observation is bundled into an accounting period rather than a snapshot).
Date adjustments: SeasonallyAdjusted, Rolling30YAvg.
Financial metrics: StandardizedCurrency vs. LocalCurrency.

Don’t’s:
• Temporal rule: Do not use for simple calendar dates; only include time when signaling a mathematical smoothing or accounting window.
• Constraints: Do not use it to express additional constraints.

carojahn (changes requested):

custom_dc/custom_data.md
kmoscoe: Which property? I don't understand.

custom_dc/custom_data.md
kmoscoe: I know what you mean. But we don't have "rate" in our graph, and I don't want to define it here.

custom_dc/custom_data.md
kmoscoe: The only reason I put it in is that having slightly different wording might help with embeddings generation. But maybe not. Do you have a better description text that would be more different?

custom_dc/custom_data.md
kmoscoe: Because I believe that's how they calculated it, according to their description: 
Percentage of school-going adolescents of a specified age range who report smoking cigarettes either daily or occasionally, usually with reference to the 30 days prior to the survey. Numerator: Number of school going adolescents of a specified are range who report having smoked cigarettes either daily or occasionally. Denominator: Total number of school-going adolescents of this specified age range.
They don't seem to have distinguished the denominator by sex AFAICT.

kmoscoe: IOW, it's the percentage of female smokers among all adolescents. Not the percentage of smokers among all females.  (I believe.)

custom_dc/custom_data.md
kmoscoe: I thought domainIncludes and rangeIncludes are only properties of properties, not statvars. Yes, I should add comparisonPeriod. Let me do that.

custom_dc/custom_data.md
kmoscoe: TBH, I don't want to get too much into rules of usage here (we should publish another doc about that), but I can certainly summarize this briefly.

kmoscoe: OK, I wrote this up. PTAL

custom_dc/custom_data.md
```diff
@@ -56,13 +57,13 @@ The following sections describe the high-level conceptual work you need to do be
...
-- The thing you are measuring, known as the `populationType` in Data Commons. Often this is a `Person`, which is a commonly used population in Data Commons. But it could be something else entirely, like the beds in a hospital, the price of a commodity, Olympic medals won by a country, or the surface area of an ocean. 
-- The level at which you want to aggregate the data. Most commonly in Data Commons this is a place type such as `City`, `Country`, `AdministrativeArea1`, etc. Examples of other entity types are `Hospital`, `PublicSchool`, `Company`, `BusStation`, `Campground`, `Library` etc. 
+* The thing you are measuring, known as the `populationType` in Data Commons. Often this is a `Person`, which is a commonly used population in Data Commons. But it could be something else entirely, like the beds in a hospital, the price of a commodity, Olympic medals won by a country, or the surface area of an ocean.
```
carojahn: I don't think price of a commodity would be a population type? similar, the surface area of an ocean wouldn't be - the ocean itself would be population type?

custom_dc/custom_data.md
```diff
@@ -72,33 +73,37 @@ An _entity_ is an instance of an entity type. For example, for `PublicSchool`, b
 {: #search}
 #### Search for an existing entity / entity type
 
-Unfortunately, it is currently not possible to get a full list of entity types or entities in the Data Commons UI. To do a complete search for an entity type or entity, you need to use the REST or Python APIs. 
+Unfortunately, it is currently not possible to get a full list of entity types or entities in the Data Commons UI. To do a complete search for an entity type or entity, you need to use the REST or Python APIs.
```
carojahn: I would delete "unfortunately"

custom_dc/custom_data.md
carojahn: Like why are you only making it more english-y in the statvar dcid, vs the property value itself. I don't understand why you are pivoting from default concatenation.

custom_dc/custom_data.md
carojahn: You could express it with the numerator and denominator. It's the number of adolescents who currently smoke cigarettes and are female among all females.

custom_dc/custom_data.md
carojahn: Denominator: Total number of school-going adolescents of this specified age range. >> this is a direct quote from them?

carojahn (changes requested):

custom_dc/custom_data.md
kmoscoe: You're right. Fixed, PTAL.

custom_dc/custom_data.md
kmoscoe: Done.

custom_dc/custom_data.md
kmoscoe: I'm confused; the populationType is Adolescents. So shouldn't the DCID be Adolescents?  I think the original point was about "Smoking". But "Ratio_Smoking_Adolescents" sounds weird. (Not my fault Smoking is the defined DCID. :-))

custom_dc/custom_data.md
kmoscoe: OK, so just leave it out then?

custom_dc/custom_data.md
kmoscoe: Yes.

kmoscoe: Here: https://screenshot.googleplex.com/1gkgtj9mpsg6g

custom_dc/custom_data.md
carojahn: No just also describe the denominator

custom_dc/custom_data.md
carojahn: That is just the generic description of the numerator and denominator. And then it says "Disaggregation: Sex". From my experience, it's best practice to show disaggregated sex numbers out of a total of all sexes, because then you are not comparing apples-to-apples. You want to filter both your numerator and denominator to the disaggregation.

carojahn (changes requested):

custom_dc/custom_data.md
kmoscoe: OK, so I changed it.

custom_dc/custom_data.md
kmoscoe: OK, done.

carojahn (approved):

PR #721, opened 2026-08-03
Added description of multi-entity observations tool

kmoscoe:


mcp/run_tools.md
```diff
@@ -1,12 +1,12 @@
 ---
 layout: default
-title: Use MCP tools
+title: Use MCP
```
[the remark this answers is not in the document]
kmoscoe: I'll go back to tools.

mcp/index.md
```diff
@@ -32,25 +32,28 @@ You can also run your own MCP server locally, or in Google Cloud Platform. If yo
...
-- `get_observations`: Fetches statistical time-series data for a given variable and place. This allows queries like:
+- `get_observations`: Fetches statistical time-series data for a given variable and place. This allows queries such as:
    - "List the population of Canada since 1964."
    - "What are the current populations of China and India?"
-- `get_child_observations`: Fetches statistical time-series data for all contained-in places of a specified type. This allows queries like:
+- `get_child_observations`: Fetches statistical time-series data for all contained-in places of a specified type. This allows queries such as:
    - "Compare the life expectancy between different countries in South America."
    - "Rank-order the GDP for all countries in Eastern Europe."
+- `get_multi_entity_observations`: Fetches statistical time-series data for observations that are related to more than one entity (or place). This optimizes queries such as:
+   - "What is the current prevalence of smoking among women in China?"
```
keyurva: Hmm - not sure if these are multi entity queries. Some examples you could use:

- "What are the current rice exports from Sri Lanka to Australia?"
- "Which African countries has China provided the most financial aid to?"

mcp/index.md
```diff
@@ -32,25 +32,28 @@ You can also run your own MCP server locally, or in Google Cloud Platform. If yo
...
+   - "Which 3 countries are the top contributors to the U.N. budget?"
 
 ## Skills
 
-The MCP server has a library of packaged skills (resources) that provide expert guidance to agents on how to interact with it. There are general skills that provide global recipes for formulating tool calls and processing responses, and more detailed skills tailored to specific use cases, such as getting data for contained-in places or metrics involving multiple dimensions (beyond places).
+The MCP server has a library of packaged skills that provide expert guidance to agents on how to interact with it. There are general skills that provide global recipes for formulating tool calls and processing responses, and more detailed skills tailored to specific use cases, such as getting data for contained-in places or metrics involving multiple dimensions (beyond places).
```
keyurva: Looking at this closely - this doesn't seem right. We have skills exposed as MCP resources that are recipes / playbooks for different types of queries. And we have tool calls that execute specific tasks (like searching for indicators and fetching metadata and data)

keyurva (approved):

mcp/index.md
kmoscoe: Much better thank you!

mcp/index.md
kmoscoe: But the skills actually tell the agent how to call the tools. I'm confused about your use of "resources". My understanding is that the "resources" are the (read-only) data you are making available to the agent, in this case observations, schema, and metadata in Spanner. "Skills" are the instructions that tell the LLM how to query for those resources. I don't understand the idea of "skills exposed as resources"; this doesn't make sense to me. I've tried to rewrite this in the way that I understand these concepts -- also rewrote the definition of the multi-entity tools, which, as you clarified, really involve an explicit directional relationship between two places. PTAL.

kmoscoe: OK, I finally understood what this means. Skills is really a client-side concept, while resources is an MCP server concept. Rewritten again.

keyurva: that's perfect - thanks!

keyurva (approved):

PR #725, opened 2026-08-07
Add documentation of V2 API extensions

kmoscoe:
This revises the description of filters and chaining, and method parameters for all endpoints. It also adds several new examples for node and resolve, and simplifies the REST observation examples (but not Python).


n-h-diaz (approved):
Thanks!

PR #730, opened 2026-08-17
Add attribution page, reference to llms.txt

kmoscoe:
This PR adds a doc for required attributions on DC data. It adds the new page to llms.txt. It also adds a mention of prompting LLMs to use that file.

Note that the actual text of the attribution doc is changed because it was out of date. Also note that I will need to regenerate llms-full.txt _after_ this PR is merged.


attribution.md
miss-o-soup, line 1: This might be a bit missleading, because the reference is simply in line:
In the below example, the source and year are both in the sentence and may therefore be dropped from the citation. 

I would say:
In the below example, the reference, source and year are both placed inline in the sentence.

miss-o-soup (approved):
Looks good to me, I just had a small comment for your consideration.
Thank you! :-)

attribution.md
kmoscoe: I agree!

attribution.md
```diff
@@ -0,0 +1,49 @@
...
+
+### Citations from a document
+
+When referencing a specific data point within a written document, report, or presentation, you must include an in-text citation. The citation must include the following:
+* Source (name of the data provider / organization)
+* Data range (year or years of data)
+* (Optional) Dataset name 
+* Text “With minor processing by Data Commons”, ideally with a deep link that goes to the actual data in Data Commons
+
+In the below example, the source and year are both in the sentence and may therefore be dropped from the citation.
```
kmoscoe:
```suggestion
In the example below, the source and year are both placed inline in the sentence. 
```

assets/citation.png
premr, line 1: i do not think we should use this image because the image and the attribution are not aligned to what we asked for / expect from data providers. (Don't use AI Mode as the example). I think it's fine right now to not use an example visually.

attribution.md
```diff
@@ -0,0 +1,49 @@
...
+
+All uses of Data Commons data must be accompanied by a citation that provides a clear path for users to trace the data back to Data Commons, the data aggregator. Data Commons provides links back to the original source, associated metadata, and deeper exploration.
+
+### Citations from a public web page
+
+If you use Data Commons data in a public website, the citation should appear on the same page where the data information is presented. The citation can be either directly next to the data or on a second line after the data is presented, using a reference number or link via an icon. The citation can be displayed behind a modal pop-in via an icon or a footnote on the same page. 
+
+Here is an example in Google Search AI Mode that uses several links that go to the Data Commons place page for the relevant query:
+
+![Example of citations in Google Search AI Mode](/assets/citation.png){: width="800"}
```
premr: I don't think we should use AI mode as an example (see comment on images)

premr (changes requested):

assets/citation.png
kmoscoe: Yes, I noticed that. I'll remove this.

attribution.md
kmoscoe: Agree.

premr (approved):

